PR #7, opened 2017-03-09
EE Bearer Only Quickstart Integration tests

sebastienblanc:
This PR adds integration tests for the service-jee-jaxrs , it uses arquillian and relies on the test helper that has been added here https://github.com/keycloak/keycloak/pull/3928 
@abstractj can you review ?


abstractj:
@sebastienblanc at the README we have:

```
1. Make sure you have an Keycloak server running with an admin user in the master realm or use the provided docker image
2. Be sure to set the TestHelper.keycloakBaseUrl in the createArchive method.
3. Run mvn clean install -Pwildfly-managed
```

I don't think we're going to run our tests against `master` using the admin user right?  That's the reason why we have `quickstarts-realm.json` am I wrong? Also, shouldn't we be using the dynamic client registration service?

Anyways I tried to configure following the README file with the following setup:

```
@Deployment(testable = false)
    public static Archive<?> createTestArchive() throws IOException{
        TestsHelper.appName = "test-demo";
        TestsHelper.baseUrl = "http://localhost:8080/test-demo";
        TestsHelper.keycloakBaseUrl  = "http://localhost:8180/auth";
        TestsHelper.ImportTestRealm("admin","admin","/quickstart-realm.json");
        TestsHelper.createDirectGrantClient();
         return ShrinkWrap.create(WebArchive.class,  "test-demo.war")
                .addPackages(true, Filters.exclude(".*Test.*"),Application.class.getPackage())
                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml")
                .addAsWebInfResource(new StringAsset(TestsHelper.createClient(generateClientRepresentation())), "keycloak.json")
                .setWebXML(new File("src/main/webapp", "WEB-INF/web.xml"));

    }
```

I get the following error:

```
Results :

Tests in error:
  ArquillianTest.org.keycloak.quickstart.jaxrs.ArquillianTest » Runtime Could no...
  ArquillianTest.cleanUp:127 » BadRequest HTTP 400 Bad Request

Tests run: 2, Failures: 0, Errors: 2, Skipped: 0

```

sebastienblanc:
For the admin user and master realm : no but we just need an admin user to be able import the realm.
it's using the dynamic client registration : `TestsHelper.createClient(generateClientRepresentation())`

For your error don't you have more logs before the results of the tests are displayed ?

abstractj:
I believe we discussed it. Don't you remember that we would have a "test-admin" user with the rights to create realms? 

Here's all the details about my error: https://gist.github.com/abstractj/75db843dc8b63544f83845dac12cb825

sebastienblanc:
Yes but that test-admin user must be defined in the master realm in order to be able to create other realm I believe ? 

For your errors, do you see anything wrong in the keycloak server logs ?

PR #14, opened 2017-04-11
KEYCLOAK-4253 add json file for creating admin user

vramik:


abstractj:
@vramik @sebastienblanc I can be wrong, but I don't think we should have this. This should be part of our README files, because we don't won't to force people to use our admin. Instead, we should just provide guidance for then to edit the code and provide the administrator.

The idea behind `test-admin`, is for scenarios were people share a Keycloak instance with others and they have no control/access to the admin's credentials.

That said I believe we can do two things:

1. Provide instructions for people to change the admin's credentials to whatever they want
2. If they don't have access to admin credentials. Instruct them to import `quickstart-realm.json` and run the tests.

Thoughts?

vramik:
@abstractj I agree we don't want force people use predefined admin. Maybe I miss something here but when you want to run tests some admin in master realm has to be created, right?

sebastienblanc:
Yes, it's already part of the `generic` readme, we should just add a note saying that if they plan to run the tests, the smoothest experience is to use `admin` as username and `admin` as password but also telling they will be able to change the testcase to suit their needs.

vramik:
Closing this one: `bin/add-user-keycloak.sh -u admin -p admin` can be used in CI job

PR #16, opened 2017-04-11
KEYCLOAK-4744 updated versions in poms

vramik:


vramik:
I've updated all modules except app-springboot and service-springboot-rest. These I've left unchanged.

abstractj:
@vramik landed, thank you

PR #18, opened 2017-04-18
Set version to 3.0.0.Final

abstractj:


sebastienblanc:
Since this week we are releasing 3.1 CR and next week final , I propose we close this and do this for 3.1 / 3.2 , @abstractj wdyt  ?

stianst:
There's other things here as well as bumping version. I'm OK with merging this. For bumping to 3.1.0 I'm going to do that automatically as part of release jobs.

PR #22, opened 2017-04-27
Angular 2 Quickstart

ssilvert:


ssilvert:
The Travis log is full of garbage.  Not sure what is going on here.

abstractj:
@ssilvert I will fix Travis and move to 3.1.0.Final. Despite that, I couldn't find any integration tests in your PR. Could you please write them?

ssilvert:
Added tests.  Now tests run on Windows too.  :-)

abstractj:
@ssilvert I love this gaming OS system. Could you please rebase, Travis is already fixed on master.

preetham1290:
This provided example works fine. But if we change LocationStrategy to { provide: LocationStrategy, useClass: HashLocationStrategy }. Then there will be an issue when successfully logged in, KC redirect with URL http://localhost:4200/?redirect_fragment=%2Flogin#state=400e1361... So now angular actually tries to redirect to '#state' but such route will not be configured so throwing exception. Can we avoid using #state from KC

ssilvert:
I don't have time to take a closer look at this, but I can tell you that HashLocationStrategy will work with Keycloak.  I'm rewriting the Account management app using Angular 4 and I'm using HashLocationStrategy.
https://github.com/ssilvert/keycloak/blob/ACCOUNT-REST/themes/src/main/resources/theme/keycloak-preview/account/resources/app/app.module.ts#L92

preetham1290:
I went through your application and i was able to do exactly like you did. But still getting same error.
And i have few doubts if you can answer here or anywhere would be really very helpful.
1. Where are you invoking init function in KeycloakService?
2. Where are you invoking login function from?
I invoked init function in some component within onInit() 
And KeycloakService's login() on click of button.
Is there anything that i am missing?

abstractj:
@preetham1290 please file a Jira with exact steps to reproduce your issue or ask on the mailing list. We don't check comments on closed PRs.

Thank you

PR #25, opened 2017-05-10
App angular2

ssilvert:


app-angular2/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <parent>
+        <groupId>org.keycloak.quickstarts</groupId>
+        <artifactId>keycloak-quickstart-parent</artifactId>
+        <version>3.1.0.CR1-SNAPSHOT</version>
```
abstractj: You can bump this up to 3.1.0.Final. We already did it on master.

abstractj:
@ssilvert I bumped it up to 3.1.0.Final. But tests are still failing on Travis https://travis-ci.org/abstractj/keycloak-quickstarts/builds/230825456. I didn't have enough time to debug what's wrong. But if you have an idea, that would be nice.

ssilvert:
I bumped the version too!!

Just realized that the tests I added can not be run in Travis because we decided not to include the node_modules directory.  But I don't think that's the thing that is causing the build to fail.  Looks like it fails long before it gets to that.

abstractj:
@vmuzikar do you think is a good idea to run npm install as a build step for this module? Otherwise, tests will fail for the lack of `node_modules` folder.

vmuzikar:
Agree - imho it would make sense to run npm install during maven build.
If I understand it correctly, you need to either way run npm install before the deployment of the app, so way not to incorporate it in there.

However, since it's a quick start, I would mentioned in the README what's happening in there - that npm install is executed for you by maven.

abstractj:
Landed here https://github.com/keycloak/keycloak-quickstarts/commit/c816d7c47552843d5443b325cb7eb7c016e93b60

PR #31, opened 2017-06-21
Increase page load timeout for ArquillianAngular2Test

vramik:


vramik:
I found out that other tests also fails when html element <div id="loading" class="loading">Loading...</div>
is desplayed on the page, I'll create new PR with more complex fix, closing this one.

PR #32, opened 2017-06-22
KEYCLOAK-5081 Add more info to log when test fails

vramik:


vramik:
The PR is ready to be merged, thanks

PR #33, opened 2017-07-06
Cleanup and add profile for version bump

sebastienblanc:
@stianst : This PR removes KC parent pom and adds a new profile containing all the modules to run the `set-version` script.

### DIFF
--- a/app-springboot/pom.xml
+++ b/app-springboot/pom.xml
@@ -17,13 +17,6 @@
     <dependencyManagement>
         <dependencies>
             <dependency>
-                <groupId>org.keycloak.bom</groupId>
-                <artifactId>keycloak-adapter-bom</artifactId>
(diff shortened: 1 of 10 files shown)


stianst:
Failing on Travis as it's trying to use 3.2.0.CR1-SNAPSHOT to run tests which isn't there obviously. How should we deal with that? I guess ideally we'd like master to use KC master as that allows us to develop quickstarts for new features that are being developed. Would it be a crazy idea to have the tests checkout and build KC master?

sebastienblanc:
@stianst Yeah I did the PR without running your `set-version.sh` script (to avoid all the version changes commit) and the travis already check-out and build KC master as you suggested and for this reason only `3.3.0-CR1-SNAPSHOT` are available ;)
On this branch I ran your script https://travis-ci.org/sebastienblanc/keycloak-quickstarts/builds/250795739 and it works

PR #47, opened 2017-08-31
Set version to 3.3.0.CR1

abstractj:
I'm not sure if we already have a job to do this, but there we go.


abstractj:
Closing it, out of date.

PR #53, opened 2017-10-17
Workaround to skip Angular.js app on Travis CI

abstractj:
Our quickstarts examples are failing while running Angular.js example on
Travis CI. It's important to notice that all the tests pass locally.

The problem seems related with Webdriver while running integration tests
against Angular.js app on Travis CI. Even if we increase the timeout, it
does not work.

This change will skip Angular example ONLY on Travis.

I reckon that this is not the best solution, but while we don't have too much time and bandwidth to debug what's happening on Travis. I suggest us to merge this PR, to unblock others and file a new Jira to investigate it in further releases.


abstractj:
I just closed this, because after clean Travis' cache, this error has nothing to do with the Angular.js app

PR #58, opened 2017-10-20
Use shallow clones

abstractj:


sebastienblanc:
@abstractj Can you observe better perf with this on travis ? Anyway I think we can merge this , right ?

abstractj:
@sebastienblanc there's no huge improvement. Still, it's something nice to have. Landed here https://github.com/keycloak/keycloak-quickstarts/commit/5fb8f99493fc3049e7f5d7010550f3c98346fa01

PR #73, opened 2017-11-30
KEYCLOAK-5949 : update current repo instead of adding new one

sebastienblanc:
@vramik : this should do the trick, wdyt ?


vramik:
@sebastienblanc it seems so, yeah. How often is https://github.com/redhat-developer/redhat-sso-quickstarts/blob/7.2.x-devel actually updated. Right now there is still old version of the pom.xml

sebastienblanc:
@vramik each time that something is merged to master in the community quickstarts it should push to the redhat-dev repo as well.

sebastienblanc:
@vramik Maybe let's just merge this and see if it's okay what is pushed ?

vramik:
@sebastienblanc agreed

abstractj:
@sebastienblanc @stianst did some changes to the productize.sh script. Please, make sure that you rebased against these changes.

stianst:
I'm closing this as the productize.sh script now removes the old repositories block before adding a new one. So this issue has been resolved.

PR #75, opened 2017-12-08
Fixed a bug within the Service SpringBoot Rest example

arekmano:
- The example would return HTTP 401 Status code when
  accessing the "/products" path, but appending a "/"
  to the path allowed bypassing the auth layer.
- Example:
  - curl http://{URL}/products -> HTTP 401
  - curl http://{URL}/products/ -> HTTP 200, Body:
    ["alpha", "bravo", "charlie", "delta", "echo", "foxtrot"]
- Fixed the example by adding a new Pattern "products/"


arekmano:
Any feedback about this particular change?

matthewhelmke:
Fine from my end, we are waiting on someone from the dev/engineering team to have time to look. Pinging @stianst just in case it was missed.

stianst:
@abstractj can you review this one please?

abstractj:
@arekmano landed, thank you

PR #78, opened 2017-12-20
KEYCLOAK-5543 - Edited QS readme

Pepo48:
To preserve consistency, alternative client creation option is mentioned.

### DIFF
--- a/service-jee-jaxrs/README.md
+++ b/service-jee-jaxrs/README.md
@@ -64,6 +64,8 @@ different host. To do this edit `keycloak.json` and add:
 }
 ````
 
+As an alternative you can create the client by importing the file [client-import.json](config/client-import.json) and
+copying [config/keycloak-example.json](config/keycloak-example.json) to `config/keycloak.json`.
 
 Build and Deploy the Quickstart
 -------------------------------


abstractj:
@matthewhelmke would you like to review?

service-jee-jaxrs/README.md
```diff
@@ -64,6 +64,8 @@ different host. To do this edit `keycloak.json` and add:
 }
 ````
 
+As an alternative you can create the client by importing the file [client-import.json](config/client-import.json) and
+copying [config/keycloak-example.json](config/keycloak-example.json) to `config/keycloak.json`.
```
matthewhelmke: As an alternative, you can create the client by importing the [client-import.json](config/client-import.json) file and
copying [config/keycloak-example.json](config/keycloak-example.json) to `config/keycloak.json`.

abstractj:
@Pepo48 could you please update based on Matthew's comments?

Pepo48:
@abstractj according to suggested, it seems that similar edit could be done in another 4 QSs, at least (I have also just copied this sentence from another QS). Also, because of fact that this PR is an "orphan" (I have deleted an original fork), I would suggest close this one and create new, with @matthewhelmke  change in every other QS readme that contains the same sentence. Do you agree?

abstractj:
@Pepo48 +1 closing it, thank you.

PR #79, opened 2017-12-20
KEYCLOAK-5531 - Edited keycloak.json

Pepo48:
Fixed inconsistency between config files - client-import.json
and keycloak.json, which produces an error during deployment.


pedroigor:
@Pepo48 , can we close this one ? I think we already have this and other changes in this PR [1]

[1] https://github.com/keycloak/keycloak-quickstarts/pull/80

Pepo48:
According to changes in #80, of course, this can be closed. Thanks @pedroigor

PR #84, opened 2018-01-04
Keycloak 6117

stianst:


abstractj:
@stianst if we aim to have this on master, I believe the correct should be `4.0.0.CR1-SNAPSHOT`. Travis script will clone Keycloak repository and obtain the latest version from [here](https://github.com/keycloak/keycloak/blob/master/pom.xml#L36). Otherwise, if the idea is to target `3.4.3.Final`, I suggest to remove the -SNAPSHOT suffix and submit these changes against `latest`. This branch uses only stable and already released versions.

stianst:
@abstractj we first need to get the quickstarts finished for RHSSO 7.2. Once that's done master can move on to 4.0.

PR #85, opened 2018-01-05
[KEYCLOAK-6141] Quickstart build for 'latest' branch is broken

abstractj:


stianst:
When we next do a KC release these changes would be lost unless we apply to master branch as well. The changes to travis-server.sh I guess should just be applied to master as well, but the changes to remove group6/productize.sh would break the syncing to RHSSO repo. Rather than delete the execution of productize.sh should we not check the branch name and only run of it's the master branch?

abstractj:
+1 I'm closing this PR

PR #95, opened 2018-01-26
KEYCLOAK-6377 Allow dry run without pushing

stianst:


sebastienblanc:
:+1:

stianst:
Rebased

abstractj:
@stianst if we run the same script twice we get an error from git that `prod_staging` already exists. Maybe we should always delete this branch before run the script?

sebastienblanc:
@abstractj yes we can add this for extra safety even if this script is supposed to run only on travis and implies that the repo don't exist at the beginning of each run.

stianst:
Added a warning if branch exists. Deleting branch is tricky as the user may or may not be on that branch. If the user is on that branch we don't know what branch to switch to (it's not going to be master in most cases, it'll be some branch for a KC JIRA).

PR #102, opened 2018-03-16
[KEYCLOAK-6543] Make the photoz example a quickstart.

sguilhen:
We need to port the missing examples as quickstarts so that the authz examples in Keycloak can be deleted.


pedroigor:
@abstractj, can we have this one in ? Upstream changes already merged.

sguilhen:
@abstractj working on it, thanks for the review

pedroigor:
@abstractj, The tests now are implemented and I have successfully run them after Stefan's latest changes to this PR.

However, I think Travis won't be able to run these tests in particular because they need either FF or Chrome. Pretty much the same "problem" we have in upstream, where some tests don't work with phamtonjs and are not run in Travis.

What do you think ? Can we start merging this one ?

pedroigor:
@abstractj, OK. Will check if Stefan can do that. Otherwise I'll send another PR with his changes + the one you requested.

pedroigor:
Closed in favor of https://github.com/keycloak/keycloak-quickstarts/pull/104.

PR #109, opened 2018-05-02
[KEYCLOAK-7285] Turn off cache on Travis for the quickstarts

abstractj:


stianst:
Looks like Travis is unhappy again

PR #111, opened 2018-05-08
Fixing URLs to point to Keycloak, not Widfly

okrasz:
The Keycloak is available under /auth, not at root. The links were pointing to Wildfly console, instead of Keycloak.


stianst:
Can you please send PR to master branch? Latest branch is just synced from master at release time. Also, please don't update the versions.

PR #116, opened 2018-06-19
Keycloak 7625/enforcer cleanup 2

slaskawi:
https://issues.jboss.org/browse/KEYCLOAK-7625

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,10 +23,11 @@ env:
 
 before_install:
   - ./travis-server.sh
+  - ./scripts/generate_keycloak_json.sh
 
 install:
-  - travis_wait 20 mvn -s maven-settings.xml clean install --no-snapshot-updates -Denforcer.skip=true -B -V -q
-  - mvn -s maven-settings.xml -Padd-datasource install -Denforcer.skip=true -f user-storage-jpa -q
+  - travis_wait 20 mvn -s maven-settings.xml clean install --no-snapshot-updates -B -V -q
+  - mvn -s maven-settings.xml -Padd-datasource install -f user-storage-jpa -q
 
 before_script:
   - export MAVEN_SKIP_RC=true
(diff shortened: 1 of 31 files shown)


slaskawi:
Looking at the [build history](https://travis-ci.org/keycloak/keycloak-quickstarts/builds), I guess we'll need some stabilization work to be done. I'm on it!

abstractj:
@slaskawi would you mind squash your commits in a single one?

slaskawi:
@abstractj Squashed.

slaskawi:
@abstractj Rebased.

abstractj:
@slaskawi if we run the following steps against master:

```
$ git clone https://github.com/keycloak/keycloak-quickstarts
$ cd keycloak-quickstarts/app-profile-jee-vanilla
$ mvn clean wildfly:deploy
```
The application is deployed with no issues, even without your patch. Now I'm confused which issue this PR is fixing. Could you please walk me through the steps to reproduce?

slaskawi:
@abstractj I believe there is a mistake in the ticket description. If you look at the comment, you will notice that zhengkai invoked `mvn` command in the project's root (not the `app-profile-jee-vanilla` directory). If you do that, you will probably notice that enforcer fails (unless you use `-Denforcer.skip=true`).

When investigating this, I noticed that we we had lots and lots of duplicated enforcer plugin definitions all over the place. That's why I though, it might be a good idea to unify them. Now the plugin configuration is done in parent's `/build/pluginManagement` section and all you need to do is to run it in the child's pom. This way, the configuration is less error prone and more consistent.

The final piece of this commit is adding `scripts/generate_keycloak_json.sh` which creates `keycloak.json` files from `keycloak-example.json`. This allowed me to run Travis *with* enforcer plugin (so far it has been skipped during the Travis build, which was a bit weird).

slaskawi:
@abstractj Rebased & repushed

PR #122, opened 2018-08-01
Set version to 4.3.0.Final-SNAPSHOT

mposolda:


abstractj:
@mposolda I believe this PR could be squashed in a single commit. What do you think?

mposolda:
@abstractj yeah, ok. I've squashed to single commit. I've originally used separate commit for adding of modules to pom.xml and changed the version, but if you prefer single commit, it's fine with me.

abstractj:
@mposolda closing this and replacing it by https://github.com/keycloak/keycloak-quickstarts/pull/124

PR #129, opened 2018-10-17
KEYCLOAK-8543 Authorization quickstarts won't work on wf14

vramik:


abstractj:
@mposolda @vramik correct me if I'm wrong. But I believe we need more than this change for WF 14. Actually we're targeting `WildFly 11.0.0.Final` and the BOM file `wildfly-javaee7`. I believe it has to be changed, am I wrong?

vramik:
@abstractj this PR was initially just about make the quickstarts working also with wildfly14. Making wildlfy 14 as default wildfly version is out of scope of this. Is there a requirement to upgrade to wf14? If so, we can do it either within another issue or within this one (and update the description in jira). wdyt?

mposolda:
@abstractj @vramik I vote for merge this as authz quickstarts will be broken on wf14 without it and we know that other wf/eap versions also work fine with this.

Then fixing the build (and upgrade to newer wf?) can be likely follow-up task? The question is, if broken quickstart build is real issue signalizing the fact that quickstarts are really broken, or if it's just issue with the tests/stability. Do we know this? If it's the real issue, then should be likely fixed before 4.6.0 release?

abstractj:
> @abstractj @vramik I vote for merge this as authz quickstarts will be broken on wf14 without it and we know that other wf/eap versions also work fine with this.

@mposolda if you are comfortable with this, go for it. I'm removing myself from the review to not be a blocker to you guys.

> 
> Then fixing the build (and upgrade to newer wf?) can be likely follow-up task? The question is, if broken quickstart build is real issue signalizing the fact that quickstarts are really broken, or if it's just issue with the tests/stability. Do we know this? If it's the real issue, then should be likely fixed before 4.6.0 release?

@mposolda @vramik If you look at the Travis logs output, it seems like things are badly broken. Whether or not the upgrade is gonna fix it, honestly, I don't know. IMO it's hard to know if there are real issues, with the build broken. Anyways, feel free to use your best judgment.

abstractj:
@vramik @mposolda superseded by https://github.com/keycloak/keycloak-quickstarts/pull/134. Rebased against master and merged. Will try to stabilize the build after this.

PR #130, opened 2018-10-23
Correct basic errors in docs

melissapalmer:
- for spelling
- added more links to other quickstarts on main README


stianst:
Changes look good, but this is the wrong branch. latest branch is updated when we do Keycloak release, any changes should be done to master branch. Can you please open the same PR against master branch?

melissapalmer:
Ok, will open against master branch

PR #133, opened 2018-11-04
WildFly 15 and java 11 upgrade

flashboss:
I have an upgraded version working on jdk 11 and the first WildFly 15 alpha version. I also updated the other libraries using more recent ones. Removed some warning in compile and sorted the maven variables in a more unique way. All tests with different profiles work. I hope it can be useful

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,8 +129,8 @@ The next step is to start <span>WildFly</span> server:
 1. Open a terminal and navigate to the root of the <span>WildFly</span> server directory.
 2. Use the following command to start the <span>WildFly</span> server:
    ````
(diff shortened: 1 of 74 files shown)


stianst:
Hi, thanks for your PR, but there are a few issues here. The PR is to the wrong branch. "latest" is to latest released version of the QuickStarts. Development goes into "master". WildFly 15 has not been released yet and as such we are not targeting this yet. We also still want to keep support for Java 10 and not upgrade to Java 11 just yet. As there are so many changes in one single PR it's not possible for us to review/handle this PR. Would be better if you break it up and also create JIRAs for individual changes.

PR #142, opened 2018-11-29
Correct error of quickstart example user-storage-jpa

deelthor:
Example wont work because the jta-datasource config in the persistence.xml did not reference the XA datasource defined in the pom.xml. It was possible though to deploy the user storage to the Keycloak instance and to load it. But when creating a user you would see the error that the JDBC Connection could not be established.

### DIFF
--- a/action-token-authenticator/pom.xml
+++ b/action-token-authenticator/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>keycloak-quickstart-parent</artifactId>
         <groupId>org.keycloak.quickstarts</groupId>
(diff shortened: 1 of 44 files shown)


abstractj:
@deelthor we really appreciate the time it took to create this submission. Unfortunately, we were not able to review last year. 
I would like to ask you to file a Jira (https://www.keycloak.org/community.html) describing the issue reported, steps to reproduce and affected version. In this way, we can prioritize and give the correct attention to this change.

abstractj:
@deelthor I'm closing this due to the lack of a proper Jira and also this PR needs a rebase. Please, do not hesitate to reopen when you think it's ready.

PR #150, opened 2019-03-07
Remove deprecated install command from service-nodejs's package.json

antonioalmeida:
The ```service-nodejs``` example can't be installed because of the deprecated install script ```nsp check``` .

Removing it from the ```package.json``` does the job.
Also updated the running instructions.

### DIFF
--- a/service-nodejs/README.md
+++ b/service-nodejs/README.md
@@ -64,11 +64,11 @@ Build and Deploy the Quickstart
 
 1. Open a terminal and navigate to the root directory of this quickstart.
 
-2. The following shows the command to run the quickstart:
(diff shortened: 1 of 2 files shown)


abstractj:
Hi @antonioalmeida, please file a Jira so we can keep track of this and prioritize for further releases. See: https://www.keycloak.org/community.html

abstractj:
@antonioalmeida it has been a while since we requested changes to this PR. While the submission is appreciated, we are not able to merge it without these changes. Because it has been so long, we are going to close the issue. Feel free to reopen with the changes included, if your time permits.

PR #155, opened 2019-10-09
Update Jackson

stianst:


stianst:
@abstractj if you could fix it that would be awesome - I have to many things going on right now, so won't be able to look at this right now

stianst:
We need to update due to known CVE's in older version. Keycloak 8 will be upgraded to 2.10.0.

PR #156, opened 2019-11-11
[KEYCLOAK-5130] - Stateless Spring Security Web App

pedroigor:


pedroigor:
The failures seem to be related to https://issues.jboss.org/browse/KEYCLOAK-11282. As a result, I've added the necessary configuration to make it work. However, we should probably keep an eye on that JIRA to check whether or not this is an issue or something we should improve for SB.

abstractj:
@stianst should we remove the hold label now?

stianst:
IMO we should not add this QuickStart for two reasons. Firstly we want to migrate away from our own Spring adapters. Secondly we should not go over board with the amount of quickstarts we have.

pedroigor:
@stianst One of the main reasons behind this quickstart is that, as you know, we lack proper integration tests for SB. And unfortunately, we use the SB related quickstart as a health check for our Spring adapters. See https://issues.redhat.com/browse/KEYCLOAK-5130.

Currently, SB quickstarts are our only source of truth for non-breaking changes in SB. Although not yet 100%.

I'm OK if you want to close this one if our plan is to stop making changes to our spring adapters.

stianst:
@pedroigor do we have no automated tests for Spring Boot at all? At least for other adapters we do have integration tests in the testsuite that includes sample applications. We really shouldn't create QuickStarts as a way to test things, as creating a sample application for testing purposes is a completely different thing to creating a sample application as a starting point for users.

pedroigor:
@stianst We don't have. I'm going to close this one considering that we don't have any plans to write a better testsuite for SB.

PR #165, opened 2020-01-27
KEYCLOAK-12676 Shell script from the Quickstarts return an empty vers…

drichtarik:
…ion of Keycloak

@abstractj This version of the command checks for one or two major numbers of the version in the list for future xy.x.x Keycloak releases (eg 10.0.0) followed by "/" slash char eliminating the ".Final" string from the previous releases.


abstractj:
@drichtarik thank you, tested it locally and it's working nicely.

PR #172, opened 2020-03-16
[KEYCLOAK-11282] - Spring Boot Multi-tenancy Quickstart

pedroigor:


mhajas (dismissed):
Thanks @pedroigor, great work! It works as expected with changes from https://github.com/keycloak/keycloak/pull/6898

pedroigor:
@stianst @mhajas It seems that only a subset of spring quickstarts tests is running in CI, is that right?

vmuzikar (changes requested):
@pedroigor Could you please rebase, so we could try to run the tests with GH Actions?

pedroigor:
@vmuzikar There we go. Rebased and tests running fine locally.

vmuzikar (changes requested):
@pedroigor Thanks! That was fast. :)

Could you please add the tests to [GH Actions](https://github.com/keycloak/keycloak-quickstarts/blob/master/.github/scripts/run-tests.sh)?

pedroigor:
@vmuzikar Done. Going to cancel the previous build.

vmuzikar (approved):
@pedroigor Thanks!

The tests failed but on some unrelated QS (action-token-required-action) and some weird random error:
```
[ERROR] Failed to execute goal org.wildfly.plugins:wildfly-maven-plugin:1.2.2.Final:execute-commands (add-spi-configurations) on project provider-action-token-required-action: Execution add-spi-configurations of goal org.wildfly.plugins:wildfly-maven-plugin:1.2.2.Final:execute-commands failed: Command execution failed for command ':reload'. Communication error: java.util.concurrent.CancellationException: Operation was cancelled
```
I'd re-run the checks but I don't have the permissions. But given it was unrelated failure, I'm approving this (based on @mhajas review). I think we're good to go.

vmuzikar (changes requested):
On the other hand, the log is completely swamped with htmlunit errors `DefaultCssErrorHandler` which is weird and makes it hard to read. Any idea what's causing it?

vmuzikar:
@pedroigor Could you  please rebase? The GH Actions issues should be resolved now.

Regarding the swamped log with htmlunit warnings. It seems like it doesn't like one of the CSS. IMHO we could easily switch to chromeheadless to solve the problem. Just don't forget to add `pageLoadTimeout(30, TimeUnit.SECONDS)`, similarly to e.g. [here](https://github.com/keycloak/keycloak-quickstarts/blob/cdb5bc08af63e84d5c39d503154e08fadbbbfa54/app-jee-html5/src/test/java/org/keycloak/quickstart/ArquillianJeeHtml5Test.java#L118).

pedroigor:
@vmuzikar We don't use arquillian in this quickstart.

vmuzikar:
@pedroigor Ah, right. Ok, so maybe we could then use `webClient.setCssErrorHandler(new SilentCssErrorHandler());`? This should get rid of the trillion warn messages in the log. :) I'd really like to have the log without the messages. It's then soooo big than it even doesn't display in the GH Actions log.

pedroigor:
@vmuzikar Done. Thanks for the tip.

abstractj (approved):

PR #184, opened 2020-04-03
KEYCLOAK-13712 Keycloak Operator Quickstart Guide

slaskawi:
https://issues.redhat.com/browse/KEYCLOAK-13712

Getting Started Guide might be found here: https://github.com/keycloak/keycloak-web/pull/124

### DIFF
--- a/operator-examples/myrealm.yaml
+++ b/operator-examples/myrealm.yaml
@@ -1,0 +1,14 @@
+apiVersion: keycloak.org/v1alpha1
+kind: KeycloakRealm
+metadata:
+  name: myrealm
+  labels:
+    realm: myrealm
+spec:
+  realm:
+    realm: "myrealm"
+    enabled: True
+    displayName: "My Realm"
+  instanceSelector:
+    matchLabels:
+      app: mykeycloak
--- a/operator-examples/myuser.yaml
+++ b/operator-examples/myuser.yaml
@@ -1,0 +1,19 @@
+apiVersion: keycloak.org/v1alpha1
+kind: KeycloakUser
+metadata:
+  name: myuser
+  labels:
+    app: mykeycloak
+spec:
+  user:
+    username: "myuser"
+    firstName: "John"
+    lastName: "Doe"
+    email: "myuser@keycloak.org"
+    enabled: True
+    credentials:
+      - type: "password"
+        value: "12345"
+  realmSelector:
+    matchLabels:
+      realm: myrealm
--- a/operator-examples/README.asciidoc
+++ b/operator-examples/README.asciidoc
@@ -1,0 +1,5 @@
+Keycloak Operator Examples
+==========================
+
+This directory contains files necessary for the Keycloak Operator Getting Started Guide. Please navigate to https://www.keycloak.org/ for more information.
+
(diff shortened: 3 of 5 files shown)


operator-examples/myrealm.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: keycloak.org/v1alpha1
+kind: KeycloakRealm
+metadata:
+  name: myrealm
+  labels:
+    realm: myrealm
+spec:
+  realm:
+    id: "myrealm"
```
stianst, line 9: Why are you setting "id" here? This is really an internal ID and shouldn't be set here.

operator-examples/myrealm.yaml
```diff
@@ -0,0 +1,15 @@
...
+    realm: "myrealm"
+    enabled: True
+    displayName: "myrealm"
```
stianst, line 12:
```suggestion
    displayName: "My Realm"
```

operator-examples/myrealm.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: KeycloakRealm
+metadata:
+  name: myrealm
+  labels:
+    realm: myrealm
+spec:
+  realm:
+    id: "myrealm"
+    realm: "myrealm"
+    enabled: True
```
stianst, line 11: A shame this is needed, as it really should be enabled by default ;)

operator-examples/keycloak.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: keycloak.org/v1alpha1
+kind: Keycloak
+metadata:
+  name: mykeycloak
+  labels:
+    app: sso
```
stianst, line 6: Would it be a bit more clear if the label was something more like "mykeycloak"?

operator-examples/myuser.yaml
```diff
@@ -0,0 +1,20 @@
...
+  labels:
+    app: sso
+spec:
+  user:
+    username: "myuser"
+    firstName: "John"
+    lastName: "Doe"
+    email: "myuser@keycloak.org"
+    enabled: True
+    emailVerified: False
```
stianst, line 14: I'd remove this

operator-examples/myuser.yaml
```diff
@@ -0,0 +1,20 @@
...
+    credentials:
+      - type: "password"
+        value: "12345"
```
stianst, line 17: There's something a bit odd about having credentials in clear-text here. It's fine for now, but I think this means we need to consider how/if users are provisioned as CRs.

operator-examples/myrealm.yaml
slaskawi: Removed.

operator-examples/myrealm.yaml
slaskawi: Fixed.

operator-examples/keycloak.yaml
slaskawi: Updated.

operator-examples/myuser.yaml
slaskawi: Done

operator-examples/myuser.yaml
slaskawi: +1 for the future consideration.

operator-examples/myrealm.yaml
slaskawi: +1

PR #192, opened 2020-04-27
KEYCLOAK-13900 Wrong port in user-storage-jpa quickstart

mabartos:
JIRA : [KEYCLOAK-13900](https://issues.redhat.com/browse/KEYCLOAK-13900)


sguilhen:
@stianst could you please review this?

user-storage-jpa/src/main/resources/META-INF/persistence.xml
```diff
@@ -5,7 +5,7 @@
         http://java.sun.com/xml/ns/persistence
         http://java.sun.com/xml/ns/persistence/persistence_2_0.xsd">
     <persistence-unit name="user-storage-jpa-example">
-        <jta-data-source>java:jboss/datasources/ExampleDS</jta-data-source>
+        <jta-data-source>java:jboss/datasources/ExampleXADS</jta-data-source>
```
stianst, line 8: There's not an ExampleXADS registered is there?

mabartos: AFAIK the _ExampleXADS_ is registered, when the user add datasource by maven profile `add-datasource`. Everything is described in README. This datasource is related only for this QS, because it has to be a `xa` datasource and _KeycloakDS_ is default `non-xa` DS and ExampleDS is `non-xa` too.

mabartos:
Wrong branch; PR will be redirected.

PR #197, opened 2020-04-30
KEYCLOAK-13900 Wrong port in user-storage-jpa quickstart

mabartos:
JIRA: [KEYCLOAK-13900](https://issues.redhat.com/browse/KEYCLOAK-13900)

AFAIK the _ExampleXADS_ is registered, when the user add datasource by maven profile `add-datasource`. This datasource is related only for this QS, because it has to be a `xa datasource` and _KeycloakDS_ is default `non-xa DS` and _ExampleDS_ is `non-xa` too.


stianst:
@mabartos thanks for the explanation

PR #199, opened 2020-05-13
Define source&target compiler to prevent compilation errors while con…

muhammedyalcin:
…sidering default version.

Javac version needs to be define in order to build examples.


abstractj:
@muhammedyalcin thanks for submitting this, but could you please read https://github.com/keycloak/keycloak-quickstarts/blob/latest/CONTRIBUTING.md? As you noticed the build is failing, plus we need a Jira with all the details about the issue.

vmuzikar:
@muhammedyalcin Thank you for your contribution! I believe this was already fixed in #243.

@abstractj Can you please close this one?

PR #201, opened 2020-07-01
[KEYCLOAK-13869] Upgrade dependencies for service-nodejs

abstractj:
## Summary

The Node.js quickstart had a set of unused dependencies that were outdated and started to be reported by npm audit. Most of them are just utilities which we don't use, like Coveralls.

This change removes all the dependencies which we don't need and also includes a new task (`npm test`), so people could run the tests under the `test/` folder.

### DIFF
--- a/service-nodejs/package-lock.json
+++ b/service-nodejs/package-lock.json
@@ -1,0 +1,3036 @@
+{
+  "name": "service-nodejs",
+  "version": "0.0.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/parser": {
+      "version": "7.8.3",
+      "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.8.3.tgz",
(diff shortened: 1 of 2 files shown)


edewit:
Guess it would be nice to also fix the build:
```
Some problems were encountered while building the effective model for org.keycloak:keycloak-parent:pom:11.0.0-SNAPSHOT
[WARNING] 'parent.relativePath' of POM org.keycloak:keycloak-parent:11.0.0-SNAPSHOT (/home/travis/build/keycloak/keycloak-quickstarts/keycloak/pom.xml) points at org.keycloak.quickstarts:keycloak-quickstart-parent instead of org.jboss:jboss-parent, please verify your project structure @ line 24, column 13
```

abstractj:
@edewit even tough it's a good idea, that's out of the scope for this PR. But I'd appreciate if you could create a new Jira and assign to team-cheetah if you would like.

PR #222, opened 2021-01-11
KEYCLOAK-16490 Add org.eclipse.yasson dependency to spring qs and upd…

mhajas:
…ate version to 1.0.8


mhajas:
This PR is sent to wrong branch. Replaced by: https://github.com/keycloak/keycloak-quickstarts/pull/223

PR #226, opened 2021-02-08
KEYCLOAK-16487 KEYCLOAK-16505 Fix photoz QS

vmuzikar:
* Fixes both `app-authz-photoz` and `app-authz-uma-photoz` Quickstarts (both test and the QS were broken). Fixes are very similar for both QS, that's why I put them into a single PR.
* Switched to Chrome headless as it's more robust in tests than FF and also to align it with other QS.
* Updated the READMEs to reflect the new Account Console.
* Tests now uses Account REST API instead of UI not to rely on Account Console (which is out of scope of the tests here). This makes it more robust and stable.

### DIFF
--- a/app-authz-photoz/photoz-restful-api/src/main/resources/photoz-restful-api-authz-service.json
+++ b/app-authz-photoz/photoz-restful-api/src/main/resources/photoz-restful-api-authz-service.json
@@ -115,7 +115,6 @@
       "logic": "POSITIVE",
       "decisionStrategy": "UNANIMOUS",
       "config": {
-        "defaultResourceType": "[\"http://photoz.com/album\"]",
         "scopes": "[\"album:view\",\"album:delete\"]",
         "applyPolicies": "[\"Only Owner and Administrators Policy\"]"
       }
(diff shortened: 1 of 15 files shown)


vmuzikar:
@pedroigor Could you please review?

app-authz-photoz/photoz-restful-api/src/main/resources/photoz-restful-api-authz-service.json
```diff
@@ -115,7 +115,6 @@
       "logic": "POSITIVE",
       "decisionStrategy": "UNANIMOUS",
       "config": {
-        "defaultResourceType": "[\"http://photoz.com/album\"]",
```
vmuzikar, line 118: Without this line removed, the permission doesn't seem to take any effect. Not sure why but since this field doesn't seem to be exposed via UI, I think it's safe to remove. Or is it a bug in KC? @pedroigor?

pedroigor: Probably not a bug because we have several tests in our test suite with higher coverage than what we do here. Including the same application.

In theory, this line should impact how permissions are evaluated due to the fact that you are defining permission for resources with a specific type. Without it, the server won't be able to enforce access because permissions are not going to match newly created resources. In this case, albums.

Let me take a look.

pedroigor: @vmuzikar It is OK. The permission changed to a be a `scope` permission, so that attribute does not make sense.

pedroigor (approved):

pedroigor:
I don't mind not using the UI during tests because that is already being done in our test suite (for the old account console).

PR #225, opened 2021-01-19
KEYCLOAK-16154 Migrate QuickStarts to GitHub Actions

vmuzikar:
I've also sneaked in one tiny fix  for `user-storage-simple` tests.

Evidence of a running job: https://github.com/abstractj/keycloak-quickstarts/actions/runs/498430706

PR #227, opened 2021-02-09
KEYCLOAK-16486 Update app-angular2 QS

mhajas:
- newer Keycloak JS adapter
- using chrome headless in testsuite


app-angular2/src/main/webapp/package.json
```diff
@@ -20,20 +20,21 @@
     "@angular/router": "~4.0.0",
     "angular-in-memory-web-api": "~0.3.0",
     "core-js": "^2.4.1",
-    "rxjs": "5.0.1",
+    "rxjs": "5.4.2",
     "systemjs": "0.19.40",
-    "zone.js": "^0.8.4"
+    "zone.js": "^0.8.4",
+    "keycloak-js": "~12.0.2"
```
vmuzikar, line 26: I wonder, couldn't we use `keycloak.js` from the server instead of hardcoding a preset version here?

vmuzikar (changes requested):
@mhajas Sorry for late review. Changes look good, added one comment, but I'm no JS/Angular expert. @edewit Could you please give this a quick look too?

Could you please also update the README with instructions on how to run the tests, similarly to e.g. [here](https://github.com/keycloak/keycloak-quickstarts/blob/latest/user-storage-simple/README.md#integration-test-of-the-quickstart)?

vmuzikar:
@mhajas One last question. You might have already mentioned this somewhere but did you try to run it with GH Actions (#225)?

mhajas:
@vmuzikar, I would say I did, but I am not sure. We can wait for it to be merged and then rebase this to be sure, WDYT?

vmuzikar:
> @vmuzikar, I would say I did, but I am not sure. We can wait for it to be merged and then rebase this to be sure, WDYT?

@mhajas Ok, works for me.

app-angular2/src/main/webapp/package.json
mhajas: I somehow missed your comment, sorry. Well I am not sure it is a good idea. I am not really an expert in angular, but I believe this is the way how the community will use it. Maybe we should change it to `latest` so that we always test with latest realease of keycloak-js. WDYT?

vmuzikar: Ok, I see your point. Even though it might not be ideal (let's say we're testing a PR to the KC repo which makes a breaking change to the adapter, this solution wouldn't catch it), let's use `latest`.

.github/scripts/export-keycloak-version.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+if [ ! -d "keycloak-dist" ]; then
+  exit 0
+fi
+
+if [ -n "$GITHUB_BASE_REF" ] && [[ "$GITHUB_BASE_REF" == "latest" ]]; then
+  export KEYCLOAK_VERSION=$(grep -oPm1 "(?<=<version>)[^<]+" pom.xml)
```
vmuzikar, line 7: I think we could use a similar command as below:
```suggestion
  export KEYCLOAK_VERSION=$(mvn -q -Dexec.executable=echo -Dexec.args='${project.version}' --non-recursive exec:exec)
```

vmuzikar (commented):
@mhajas Thanks for the update! Added one last nitpick. :)

I think it might even work with https://github.com/keycloak/keycloak/pull/7929 (I updated it to checkout KC repo to a dir with the same name as here).

.github/scripts/export-keycloak-version.sh
mhajas: Done

vmuzikar (approved):
@mhajas Thanks! Now let's wait for a review from @ssilvert or @edewit. :)

mhajas:
Thanks @vmuzikar!

edewit:
I'll leave this for @ssilvert to try out I'm a bit out of my depth on it

abstractj (approved):

PR #230, opened 2021-03-19
Adding default H2 db

bitvilag:
I had to specify a DB as it failed to start keycloak without it (on microk8s 1.19)

No JIRA issue on this topic


vmuzikar (commented):
@bitvilag Thank you for your contribution!

Specifying `DB_VENDOR` [should not be necessary](https://github.com/keycloak/keycloak-containers/blob/master/server/README.md#database). If it is, then it's a bug and the fix should go to https://github.com/keycloak/keycloak-containers.

Could you please create a JIRA with more details, like e.g. what KC image tag did you use etc.?

@abstractj Can you please close this?

PR #234, opened 2021-03-30
[KEYCLOAK-16198] Fix productize.sh to convert links correctly

sguilhen:
Also fixes links in several READMEs so that they can be properly processed by the script

### DIFF
--- a/app-authz-rest-employee/README.md
+++ b/app-authz-rest-employee/README.md
@@ -5,7 +5,7 @@ Level: Beginner
 Technologies: SpringBoot
 Summary: SpringBoot REST Service Protected Using Keycloak Authorization Services
 Target Product: Keycloak
-Source: <https://github.com/keycloak/Keycloak-quickstarts>
+Source: <https://github.com/keycloak/keycloak-quickstarts>
 
 
 What is it?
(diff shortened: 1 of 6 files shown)


vmuzikar (changes requested):
@sguilhen Is this still relevant? If it is, could you please resolve the changes?

Also, how do I test the changes? :)

sguilhen:
Hi @vmuzikar, yes it is still relevant - if anyone runs the script as it is, it will produce wrong links in the quickstarts migrated to redhat-developer. I've updated the PR.

For testing, I personally just run the script without setting any of the env variables (like GH_USER_NAME), which triggers a dry-run - it will produce a prod_staging branch but won't push it to redhat-developer. So you can inspect the produced branch and check the links have been processed and are properly pointing to the redhat-developer repo instead of the upstream repo.

Note that for this test run you need to change the GH_BOM_BRANCH var to 7.4.x as the bom for 7.5.x does not yet exist.

vmuzikar (approved):
@sguilhen As far as I can tell, looks good to me. Thanks!

Just out of curiosity, why are we excluding some QS from prod? I don't think e.g. Events SPI or Action Tokens SPI is unsupported.

abstractj (approved):

sguilhen:
@vmuzikar Sorry for the late reply. I don't honestly know if these should continue to be excluded from prod. I'll have a discussion with Hynek later this week and depending on the outcome I might be sending another PR with further fixes.

PR #237, opened 2021-04-07
KEYCLOAK-16532 Tests for "user-storage-jpa" QS are broken

vmuzikar:
As part of the fix, I basically rewrote the test as it hadn't actually verify the storage provider before.


user-storage-jpa/src/main/java/org/keycloak/quickstart/storage/user/EjbExampleUserStorageProviderFactory.java
```diff
@@ -46,7 +47,7 @@ public EjbExampleUserStorageProvider create(KeycloakSession session, ComponentMo
 
     @Override
     public String getId() {
-        return "example-user-storage-jpa";
+        return PROVIDER_ID;
```
abstractj, line 50: @vmuzikar if we are not referencing PROVIDER_ID anywhere inside the code, I believe that's fine to write:
```
public String getId() {
  return "example-user-storage-jpa";
}
```
But that's up to you and not a deal breaker for me.

user-storage-jpa/src/test/java/org/keycloak/quickstart/ArquillianJpaStorageTest.java
```diff
@@ -21,54 +21,64 @@
...
-import static org.junit.Assert.fail;
 
 @RunWith(Arquillian.class)
 public class ArquillianJpaStorageTest {
 
-    private static final String KEYCLOAK_URL = "http://%s:%s/auth%s";
 
-    @Page
-    private LoginPage loginPage;
+    public static final String KEYCLOAK_URL = "http://localhost:8180/auth";
```
abstractj, line 57: @vmuzikar I don't see any harm about getting these changes in as is. But would make sense to make this URL dynamic? I guess we used to do that before.

abstractj (dismissed):
Added some comments, but I'm approving it anyway.

user-storage-jpa/src/main/java/org/keycloak/quickstart/storage/user/EjbExampleUserStorageProviderFactory.java
vmuzikar: It is referenced in the test:
https://github.com/keycloak/keycloak-quickstarts/blob/1f74955b2d0b02a5dd711c3d6670f664fb04a061/user-storage-jpa/src/test/java/org/keycloak/quickstart/ArquillianJpaStorageTest.java#L147-L149
I know code changes should not be made just for testing purposes, but I felt that it's a best practise to have the provider ID as a public constant anyway, so I made this chage. But let me know if you want me to move this constant to the test class.

user-storage-jpa/src/test/java/org/keycloak/quickstart/ArquillianJpaStorageTest.java
vmuzikar: Yes, some tests try to do that but it doesn't ever really fully work. AFAIK all the QS basically expect KC to be running at this address (not just the tests) so IMHO it is ok to hardcode it for testing purposes.

vmuzikar:
Rebased to trigger GH Actions.

hmlnarik (approved):
Looks great to me, thanks @vmuzikar! One small question but not a blocker for me to merge.

user-storage-jpa/src/test/java/org/keycloak/quickstart/ArquillianJpaStorageTest.java
```diff
@@ -21,54 +21,64 @@
...
-    public static Archive<?> createTestArchive() throws IOException {
+    public static Archive<?> createTestArchive() throws Exception {
+        // set storage
+        onlineManagementClient = ManagementClient.online(OnlineOptions
+                .standalone()
+                .hostAndPort("localhost", KEYCLOAK_MGMT_PORT)
+                .protocol(ManagementProtocol.HTTP_REMOTING)
+                .build());
+        onlineManagementClient.execute("xa-data-source add --name=ExampleXADS --driver-name=h2 --jndi-name=java:jboss/datasources/ExampleXADS --xa-datasource-properties={URL=>jdbc:h2:mem:test} --enabled=true");
+        onlineManagementClient.executeCli("reload");
```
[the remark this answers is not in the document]
vmuzikar: Thanks for the review! AFAIK the `@Deployment` annotated methods are called even before `@BeforeClass`. The datasource needs to be created before deploying the storage provider, otherwise the deployment fails.

PR #241, opened 2021-04-11
Update apiVersion to networking.k8s.io/v1 as extensions/v1beta1 has b…

Tin-Nguyen:
Updating apiVersion to networking.k8s.io/v1 in keycloak-ingress.yaml as extensions/v1beta1 has been deprecated in 1.20v

### DIFF
--- a/kubernetes-examples/keycloak-ingress.yaml
+++ b/kubernetes-examples/keycloak-ingress.yaml
@@ -15,6 +15,6 @@ spec:
             name: keycloak
             port:
               number: 8080
-        path: /
+        path: /auth
         pathType: Exact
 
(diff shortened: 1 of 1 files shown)


Tin-Nguyen:
Hi @team, anyone help to review the change? would like to save a bit time for the community

vmuzikar (commented):
@Tin-Nguyen Thank you for your contribution! Could you please create a JIRA for this and add its ID to the commit message? Also, could you squash the commits?

@slaskawi I think this change should be done also at the Operator side as well, right? AFAICT, we [use old API](https://github.com/keycloak/keycloak-operator/blob/410337e825ccc6668932d8dc31826d5bcad792a2/pkg/model/keycloak_ingress.go#L5) there too.

vmuzikar (changes requested):
One last thing. This change needs to go to `master` branch, not `latest`.

vmuzikar:
@abstractj Can you please close this. It needs to go to a different branch.

slaskawi:
+1 - please merge it against the master branch. But the change LGTM.

PR #247, opened 2021-05-06
KEYCLOAK-18022 Tests for "app-jee-html5" are unstable

vmuzikar:


vmuzikar:
I also snuck fixes for the Maven upgrade into this PR. The POMs changes might look suspicious but it can't be compiled without it. And it makes sense, the deps were really missing there. I was able to reproduce it locally too but I'm not sure why it was not failing before or why I didn't discover it during the QS fixes, TBH.

abstractj (approved):

PR #251, opened 2021-06-22
KEYCLOAK-16534 Quickstart verification set for Product and Project running on Jenkins

lhanusov:
@vmuzikar @pdrozd 

This is proposed change to verify MVN repo and Quickstarts related to product release.


vmuzikar (changes requested):
@lhanusov Thank you! Haven't yet checked the changes but GH Actions seem to be broken.

.github/scripts/run-tests.sh
```diff
@@ -29,24 +39,32 @@ print_failed_tests() {
...
+
 run_tests app-angular2 -Pwildfly-managed "$JS_VERSION_OPTION"
 
 # we need to run authz springboot tests first as they are the only ones relying on manual js-policies deployment
 # other tests deploy (and the removes) the policies automatically which then later removes even the manually deployed ones
 run_tests app-authz-rest-springboot -Pspring-boot
 run_tests app-authz-spring-security -Pspring-boot
 run_tests app-authz-springboot -Pspring-boot
-run_tests app-authz-springboot-multitenancy -Pspring-boot
+if [ ! -n "$PRODUCT" ] || [ "$PRODUCT" == "false" ]; then
```
vmuzikar, line 65: I don't like so many identical IFs here (`if [ ! -n "$PRODUCT" ] || [ "$PRODUCT" == "false" ]; then
`). :) Not sure what to do about this, though. Maybe create some list of excluded QS in prod and handle it in `run_tests`? Or at least group the excluded QS into a single IF?

.github/scripts/build-quickstarts.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash -e
 
-cp -f maven-settings.xml ~/.m2/settings.xml
+if [ -n "$PRODUCT" ] && [ "$PRODUCT" == "true" ]; then
+  args="-s $PRODUCT_MVN_SETTINGS  -Dmaven.repo.local=$PRODUCT_MVN_REPO"
+else
+  args="-s maven-settings.xml"
+fi
```
vmuzikar, line 7: Just a nitpick. I think we could set some `$DIST` var in this IF and use it later when copying the authz-js-policies. That would save us from one more ugly IF. :)

vmuzikar (commented):
Thank you, @lhanusov! Nice work!

One thing I'm wondering about is whether we should move the scripts from the `.github` dir to somewhere else (maybe `.scripts` or `.tests`?). It's no longer used by GH Actions only.

vmuzikar (changes requested):
Additionally, we need to fix the GH Actions which are still failing.

lhanusov:
@vmuzikar please, check the latest changes

vmuzikar (approved):
@lhanusov Thanks for the update! LGTM, approving.

There's one more thing though – **please DO NOT MERGE YET**. The KC GH Actions also [use the moved scripts](https://github.com/keycloak/keycloak/blob/512bcd14f76aed02c1bd668d434cc5ee59a57932/.github/workflows/ci.yml#L411-L418). Could you please create a PR to reflect the changes there as well? **It'll have to be merged together with this.**

vmuzikar:
Needs to be merged with https://github.com/keycloak/keycloak/pull/8212

PR #255, opened 2021-07-28
KEYCLOAK-18906 Wait for the test app to start

hmlnarik:

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -97,6 +97,15 @@ run_tests app-springboot -Pspring-boot
 # service-nodejs tests
 npm -C service-nodejs install
 npm -C service-nodejs start >/dev/null&
+# Wait for port 3000 to open for at most 30 seconds
+{
+    I=0
+    while ! curl -sfN -o /dev/null http://localhost:3000/service/public && [[ $I -lt 60 ]]; do
+         sleep 0.5
+         echo -n .
+         I=$[$I + 1]
+    done
+} 2>/dev/null
 if ! npm -C service-nodejs test 2>&1 | tee test-logs/service-nodejs.log; then
   tests_with_errors+=("service-nodejs")
 fi


scripts/run-tests.sh
```diff
@@ -97,6 +97,15 @@ run_tests app-springboot -Pspring-boot
 # service-nodejs tests
 npm -C service-nodejs install
 npm -C service-nodejs start >/dev/null&
+# Wait for port 3000 to open for at most 30 seconds
+{
+    I=0
+    while ! echo -n > /dev/tcp/localhost/3000 && [[ $I -lt 60 ]]; do
```
stianst, line 103: This just checks if socket is open, is there a window until it's able to process requests? Might it be better to use curl or wget to check it's ready returns a valid HTTP response?

stianst (approved):
LGTM, added a comment with regards to perhaps there's a small window, but that is probably overthinking it, so happy either way

scripts/run-tests.sh
hmlnarik: Yup, that's correct. Even if the window would be negligible, curl would be a bit more reliable, so I'm pushing the new version with curl

stianst (approved):

scripts/run-tests.sh
```diff
@@ -97,6 +97,15 @@ run_tests app-springboot -Pspring-boot
...
 npm -C service-nodejs start >/dev/null&
+# Wait for port 3000 to open for at most 30 seconds
+{
+    I=0
+    while ! curl -sfN -o /dev/null http://localhost:3000/service/public && [[ $I -lt 60 ]]; do
+         sleep 0.5
+         echo -n .
+         I=$[$I + 1]
+    done
+} 2>/dev/null
```
stianst, line 108: Curl supports retry itself, so you can do:
```suggestion
# Wait for port 3000 to open for at most 30 seconds
curl --retry 60 --retry-delay 1 --retry-all-errors -sfN -o /dev/null http://localhost:3000/service/public
```

PR #262, opened 2021-09-13
[KEYCLOAK-18329] updated ingress definition for new k8s version

rbbl-dev:
new pr that hopefully matches the requirements.

ticket: https://issues.redhat.com/browse/KEYCLOAK-18329

PR #273, opened 2021-12-10
Fix: keycloak-web issue 192

andreaTP:
This is a fix for: https://github.com/keycloak/keycloak-web/issues/192

### DIFF
--- a/kubernetes-examples/keycloak-ingress.yaml
+++ b/kubernetes-examples/keycloak-ingress.yaml
@@ -1,8 +1,13 @@
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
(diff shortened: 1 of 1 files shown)


andreaTP:
Just noticed that this is similar to #262 cc. @vmuzikar

vmuzikar:
Ah, good point @andreaTP! :) Let's merge #262 then as it was first. Is the `DefaultBackend` worth leaving this PR open since we already have `Path: /`?

vmuzikar:
@andreaTP I just noticed this is created in the `latest` branch. Please see the [contribution guide on branch naming](https://github.com/keycloak/keycloak-quickstarts/blob/latest/CONTRIBUTING.md#picking-correct-branch-for-your-pr). Thanks.

PR #282, opened 2022-01-07
Add the app-nodejs-html5 quickstart as an equivalent

mhdawson:
to app-jee-html5 that Node.js developers can use.

Signed-off-by: Michael Dawson <mdawson@devrus.com>


mhdawson:
I believe there are changes the other quickstarts/docs in terms of references to this starter but I thought it might make sense to add just this part as the first step. Let me know if that is ok.

The static assets are almost a copy of the ones from app-jee-html5 except where they had to be updated to reflect the differences of running service-nodejs versus service-jee-jaxrs

In terms of testing, I think the same test should work as was used for the app-jee-html5 quickstart so I left it as that. I don't think Node.js developers need to run the test to use the quickstart so I'm thinking that would be ok.

stianst (changes requested):
Please remove the deletion of .github/ISSUE_TEMPLATE/* from the PR

stianst:
@jonkoops @edewit can one of you review this please? It looks more or less a copy of html5 quickstarts, so just packaging with Node rather than WF.

mhdawson:
@stianst I think I  figured out that there was a force push to the latest branch between when I'd forked and now, resulting in my PR having the commit that was likely removed in the force push. I removed that commit and the PR no longer has the changes to the issue templates.

app-nodejs-html5/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "app-nodejs-html5",
+  "version": "0.0.1",
+  "description": "HTML5 Service Invocation Application",
+  "main": "app.js",
+  "scripts": {
```
edewit, line 6: Maybe this would be nice?
```suggestion
  "scripts": {
    "start": "node app.js",
```

edewit (approved):

app-nodejs-html5/README.md
```diff
@@ -0,0 +1,80 @@
...
+`app.js` and replace the value of `serviceUrl`.
+
+Configuration in <span>Keycloak</span>
+-----------------------
+
+Prior to running the quickstart you need to create a client in <span>Keycloak</span> and download the installation file.
+
+The following steps show how to create the client required for this quickstart:
+
+* Open the <span>Keycloak</span> admin console
```
jonkoops, line 32: Why put 'Keycloak' in a span element? Seems like that would render the same as any other text in the readme, so perhaps it can be omitted?

```suggestion
using Keycloak. Once authenticated the application shows how to invoke a service secured with Keycloak.

System Requirements
-------------------

You need to have Node.js version 12.x or later installed.

The quickstart requires that you have the [service-nodejs](../service-nodejs/README.md) running. It assumes the
services are located at `http://localhost:3000/service`. If the services are running elsewhere you need to edit
`app.js` and replace the value of `serviceUrl`.

Configuration in Keycloak
-----------------------

Prior to running the quickstart you need to create a client in Keycloak and download the installation file.

The following steps show how to create the client required for this quickstart:

* Open the Keycloak admin console
```

app-nodejs-html5/README.md
```diff
@@ -0,0 +1,80 @@
+app-nodejs-html5: HTML5 Service Invocation Application
+===================================================
+
+Level: Beginner  
+Technologies: HTML5, JavaScript  
+Summary: HTML5 Service Invocation Application run under Express  
+Target Product: Keycloak
```
jonkoops, line 7: There is some trailing whitespace here that does not need to be there.
```suggestion
Level: Beginner
Technologies: HTML5, JavaScript
Summary: HTML5 Service Invocation Application run under Express
Target Product: Keycloak
```

app-nodejs-html5/app.js
```diff
@@ -0,0 +1,9 @@
+const express = require('express');
+const app = express();
+const port = 8080;
+
+app.use('/app-html5', express.static('src/main/webapp/'));
+
+app.listen(port, () => {
+  console.log('app-nodejs-html5 listening on port 8080');
```
jonkoops, line 8: Instead of hard-coding the port number in this string you can interpolate the existing constant.
```suggestion
  console.log(`app-nodejs-html5 listening on port ${port}`);
```

app-nodejs-html5/README.md
```diff
@@ -0,0 +1,80 @@
...
+Build and Deploy the Quickstart
+--------------------------------
+
+1. Open a terminal and navigate to the root directory of this quickstart.
+
+2. The following shows the command to deploy the quickstart:
+
+   ````
+   npm install
+   node app
```
jonkoops, line 65: I agree with @edewit's suggestion that this would be better as an NPM script.
```suggestion
   npm run start
```

app-nodejs-html5/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "app-nodejs-html5",
+  "version": "0.0.1",
+  "description": "HTML5 Service Invocation Application",
+  "main": "app.js",
```
jonkoops, line 5: To use ES module syntax you can mark set the type of this package to `module`.
```suggestion
  "main": "app.js",
  "type": "module",
```

app-nodejs-html5/src/main/webapp/app.js
```diff
@@ -0,0 +1,75 @@
...
+function notAuthenticated() {
+    document.getElementById('not-authenticated').style.display = 'block';
+    document.getElementById('authenticated').style.display = 'none';
+}
+
+function authenticated() {
+    document.getElementById('not-authenticated').style.display = 'none';
+    document.getElementById('authenticated').style.display = 'block';
+    document.getElementById('message').innerHTML = 'User: ' + keycloak.tokenParsed['preferred_username'];
+}
```
jonkoops, line 29: Since the queried elements here are static in the DOM you can store them in a variable so they only have to be queried once.
```suggestion
const notAuthenticatedElement = document.getElementById('not-authenticated');
const authenticatedElement = document.getElementById('authenticated');
const messageElement = document.getElementById('message');

function notAuthenticated() {
    notAuthenticatedElement.style.display = 'block';
    authenticatedElement.style.display = 'none';
}

function authenticated() {
    notAuthenticatedElement.style.display = 'none';
    authenticatedElement.style.display = 'block';
    messageElement.innerHTML = 'User: ' + keycloak.tokenParsed['preferred_username'];
}
```

app-nodejs-html5/src/main/webapp/app.js
```diff
@@ -0,0 +1,75 @@
...
+
+function request(endpoint) {
+    var req = function() {
+        var req = new XMLHttpRequest();
```
jonkoops, line 33: `XMLHttpRequest` is an ancient API, it would be much better to use the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) here.

app-nodejs-html5/src/main/webapp/app.js
```diff
@@ -0,0 +1,75 @@
...
+    };
+
+    if (keycloak.authenticated) {
+        keycloak.updateToken(30).success(req);
+    } else {
+        req();
+    }
+}
+
+window.onload = function () {
```
jonkoops, line 63: Instead of overwriting the `onload` property here (which might be used by other code) it's better to add this as an event listener. It is also not required to wait for the whole page to be ready (loading images and styles and such), rather you want to wait for the DOM to be ready for manipulation using the [`DOMContentLoaded` ](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event)event.

```js
function onLoad() {
  // Your initialization logic here.
}

window.addEventListener('DOMContentLoaded', onLoad);
```

app-nodejs-html5/src/main/webapp/app.js
```diff
@@ -0,0 +1,75 @@
...
+    keycloak.init({ onLoad: 'check-sso', checkLoginIframeInterval: 1 }).success(function () {
```
jonkoops, line 64: There is a new [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) based API you can use here instead.
```suggestion
    keycloak.init({ onLoad: 'check-sso', checkLoginIframeInterval: 1 }).then(function () {
```

app-nodejs-html5/src/main/webapp/index.html
```diff
@@ -0,0 +1,50 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>Keycloak Example App</title>
+
+    <link rel="stylesheet" type="text/css" href="styles.css"/>
```
jonkoops, line 23: The `type` attribute is implicitly `text/css` when `rel` is set to `stylesheet`, so it can be omitted.
```suggestion
    <link rel="stylesheet" href="styles.css"/>
```

app-nodejs-html5/src/main/webapp/index.html
```diff
@@ -0,0 +1,50 @@
...
+
+    <script src="keycloak.js"></script>
```
jonkoops, line 25: Instead of copying the entire Keycloak JS library it should be either loaded from NPM or statically from the Keycloak server.

app-nodejs-html5/src/main/webapp/index.html
```diff
@@ -0,0 +1,50 @@
+<!--
```
jonkoops, line 1: Do we need to include this in every file? Seems quite redundant.

app-nodejs-html5/src/main/webapp/index.html
```diff
@@ -0,0 +1,50 @@
...
+    <link rel="stylesheet" type="text/css" href="styles.css"/>
+
+    <script src="keycloak.js"></script>
+    <script src="app.js"></script>
+</head>
+
+<body>
+    <div class="wrapper">
+        <div id="not-authenticated" class="menu">
+            <button name="loginBtn" onclick="keycloak.login()">Login</button>
```
jonkoops, line 32: Event listeners should be added by the JavaScript included by the page (as this code now relies on globals). For example in `app.js` you could add:

```js
function onLogin() {
  keycloak.login();
}

const loginButton = document.getElementById('loginBtn');

loginBtn.addEventListener('click', onLogin);
```

jonkoops (changes requested):

app-nodejs-html5/README.md
mhdawson: @jonkoops much of that text was just copied from the Java version of the example but happy to remove the span as I'm not sure why it was needed.

app-nodejs-html5/src/main/webapp/index.html
mhdawson: @jonkoops all of the webapp code was simply copied from https://github.com/keycloak/keycloak-quickstarts/tree/latest/app-jee-html5/src/main/webapp, they should likely be consistent so maybe all of those changes should be made to both copies in 1 PR?

app-nodejs-html5/package.json
mhdawson: @jonkoops can you confirm that RH SSO prefers to use the ES module syntax over the CommonJS?  There are still some problems in using ES (for example APM) such that we've not moved most of the Node.js team assets over.

app-nodejs-html5/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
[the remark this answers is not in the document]
mhdawson:
```suggestion
```

mhdawson:
@stianst  created https://github.com/keycloak/keycloak-quickstarts/pull/301/ to replace this PR so that it's against the main branch.

@jonkoops I think I incorporated all of your suggestions except those related to the html app. What I have is a copy of what's used in the J2EE app and I'm thinking they should both be the same and be updated in the same PR instead of this one?  Does that makes sense to you?

mhdawson:
Sorry it too me so long to respond to the comments, I was pre-occupied with a family issue the last few weeks.

abstractj:
@mhdawson could you please rebase?

abstractj:
@mhdawson I noticed that https://github.com/keycloak/keycloak-quickstarts/pull/301 is mentioned as a replacement for the changes proposed here. Should we close this PR?

mhdawson:
@abstractj I'll rebase. I think this should still land (along with the rest of the ones we submitted to clean up existing doc, samples ,quickstarts). My thinking is that we should update the existing examples to include whatever the replacement client is so we still want a good starting point.

mhdawson:
So in response to your last comment we can probably just close this one and dicuss in #301, closing now.

Issue #287, opened 2022-02-01
Update integration tests for QuickStarts to use Quarkus dist


stianst:
### Description

Integration tests and GitHub actions, as well as steps within quickstarts needs to be updated to use the Quarkus distribution.

pedroigor:
Please, re-introduce tests for https://github.com/keycloak/keycloak-quickstarts/pull/302.

mposolda:
Closing as this was already done and quickstarts in the `main` branch are using Quarkus distribution now

PR #302, opened 2022-03-04
Updating user-storage-jpa quickstart for the new distribution

pedroigor:


user-storage-jpa/src/main/java/org/keycloak/quickstart/storage/user/EjbExampleUserStorageProvider.java
```diff
@@ -67,18 +63,18 @@
...
 
-    @PersistenceContext
     protected EntityManager em;
 
     protected ComponentModel model;
     protected KeycloakSession session;
 
-    public void setModel(ComponentModel model) {
-        this.model = model;
+    EjbExampleUserStorageProvider(KeycloakSession session) {
```
DGuhr, line 71: Just thinking about naming here: It hasn't much to do with EJB anymore, has it? but cant come up with a better one 😄

DGuhr (approved):
Nice! :) Apart from the naming (and thats always a hard problem, but not blocking for this iteration i think ;) ) LGTM

user-storage-jpa/src/main/java/org/keycloak/quickstart/storage/user/EjbExampleUserStorageProvider.java
pedroigor: True, I'll change this.

pedroigor: @DGuhr Updated the code as well as README.md to include a bit more details about the configuration.

DGuhr (dismissed):

ahus1 (changes requested):
I tried out the example together with the PR [enabling multiple data sources](https://github.com/keycloak/keycloak/pull/10581) and found the following things that surprised me: 

* README.md contains "Integration test of the Quickstart" that is obsolete as the integration tests don't work for this Quarkus example, yet. Consider removing that part, and also the tests from the project.
* README.md references Java EE in the section "More Information", that is probably out of date?
* I can create new users and log in with these users via the UI, still I can't search for these new users I added (neither by keyword, nor by selecting "View all users"). If this is expected, please add this to the README.md
* The database setup uses an in-memory database that will drop all data once the last connection has been closed (`jdbc:h2:mem:user-store`), even when Keycloak is still running. To avoid this, append `;DB_CLOSE_DELAY=-1` to the DB URL; this will prevent deleting the contents. Otherwise Keycloak will forget about the data after a few idle minutes.
* Mention in the README.md the minimum Keycloak version where this will work (Keycloak 18-SNAPSHOT), to prevent others from trying this with Keycloak 17.
* Please create a new issue to add an integration test for this quick-start to ensure that it continues to work with new versions of Keycloak.

user-storage-jpa-legacy/src/main/resources/META-INF/services/org.keycloak.storage.UserStorageProviderFactory
```diff
@@ -0,0 +1 @@
+org.keycloak.quickstart.storage.user.MyExampleUserStorageProviderFactory
```
ahus1, line 1: @pedroigor - this should IMHO be EjbExampleUserStorageProviderFactory; this makes the integration test of the old legacy test fail.

pedroigor: Yeah, the renaming went to far.

DGuhr:
@ahus1 thanks for trying it out! just double-checked and you are right, login / creation works (that's what i have tested) but search does not. Also not sure if it should, but guess so.

pedroigor:
@ahus1 Thanks for the review. Updated the PR with most of your suggestions. 

What is missing is bellow:

> * README.md contains "Integration test of the Quickstart" that is obsolete as the integration tests don't work for this Quarkus example, yet. Consider removing that part, and also the tests from the project.

Tests were removed for now and we have another issue to change this repo to run against the new distribution. See https://github.com/keycloak/keycloak-quickstarts/issues/287.

> * I can create new users and log in with these users via the UI, still I can't search for these new users I added (neither by keyword, nor by selecting "View all users"). If this is expected, please add this to the README.md

I'm able to query users by keyword and the behavior when querying all users and not showing those from the provider is the same as for the legacy distribution. I'm keeping the same instructions from legacy.

> * Mention in the README.md the minimum Keycloak version where this will work (Keycloak 18-SNAPSHOT), to prevent others from trying this with Keycloak 17.

I don't think we should include this as it is going to become obsolete very soon.

> * Please create a new issue to add an integration test for this quick-start to ensure that it continues to work with new versions of Keycloak.

As previously mentioned, see https://github.com/keycloak/keycloak-quickstarts/issues/287.

ahus1 (approved):
Changes are good for now and show a minimal working provider.

DGuhr (approved):

user-storage-jpa/README.md
```diff
@@ -21,45 +21,27 @@ runtime. Before using this example, you should probably read the User Storage SP
...
 
-You must first deploy the datasource it uses.
-Start up the <span>Keycloak</span> server.  Then in the directory of this example type the following maven command:
-
-   ````
-   mvn -Padd-datasource install
-   ````
+You must first configure the datasource it uses. 
+For that, copy the [conf/quarkus.properties](conf/quarkus.properties) to the `conf` directory of the server distribution.
+In this file, you have a named datasource using the same name as the [src/resources/META-INF/persistence.xml](persistence unit) from the quickstart. By using the same name,
```
dasniko, line 31: The linked resource is a) in the wrong order and b) there's a path missing
It should be: `[persistence unit](src/main/resources/META-INF/persistence.xml)`

user-storage-jpa/README.md
```diff
@@ -21,45 +21,27 @@ runtime. Before using this example, you should probably read the User Storage SP
...
+   mvn clean install
    ````
 
-You only need to execute this maven command once.  If you execute this again, then you will get an error message that the datasource
-already exists.
-
-If you open the pom.xml file you'll see that the add-datasource profile creates an XA datasource using the built
-in H2 database that comes with the server.  An XA datasource is required because you cannot use two non-xa datasources
-in the same transaction.  The <span>Keycloak</span> database is non-xa.
+To install the provider, copy the [target/user-storage-jpa-example.jar] JAR file to the `providers` directory of the server distribution.
```
dasniko, line 40: Probably you wanted to put the `target/user-storage-jpa-example.jar` in `` instead of `[]`?

user-storage-jpa/README.md
pedroigor: Done

pedroigor:
@dasniko Thanks. Applied the changes.

user-storage-jpa/conf/quarkus.properties
```diff
@@ -0,0 +1,3 @@
+quarkus.datasource.user-store.db-kind=h2
```
DGuhr, line 1: @pedroigor while I am reworking the quarkus properties behaviour i noticed that the standard transaction mode is "enabled" instead of "xa" as it is for us. I am not sure if we should advertise to set transaction mode to XA here, too ( `quarkus.datasource.user-store.transactions=xa` ), so wanted to ask if needed

DGuhr (approved):
@pedroigor @stianst afaik this one should've been merged a while ago, no? Was irritated bc of #314 but seen this one isn't merged yet. So imo we should merge.

stianst (approved):

PR #304, opened 2022-03-08
Fix keycloak.yaml

### LINKED ISSUE #300: Latest openshift-examples/keycloak.yaml of 17.0.0 does not work
### Describe the bug

In the past, using below command from https://www.keycloak.org/getting-started/getting-started-openshift :
```
$ oc process -f https://raw.githubusercontent.com/keycloak/keycloak-quickstarts/latest/openshift-examples/keycloak.yaml \
    -p KEYCLOAK_USER=admin \
    -p KEYCLOAK_PASSWORD=admin \
    -p NAMESPACE=keycloak \
| oc create -f -
service/keycloak created
route.route.openshift.io/keycloak created
deploymentconfig.apps.openshift.io/keycloak created
```
The keycloak pod can be Running. But now, it is CrashLoopBackOff:
```
$ oc get po
NAME                READY   STATUS             RESTARTS      AGE
keycloak-1-6pmv9    0/1     CrashLoopBackOff   5 (24s ago)   3m27s
keycloak-1-deploy   1/1     Running            0             3m31s
$ oc describe po keycloak-1-6pmv9
...
Events:
...
  Normal   Pulled     2m41s (x4 over 3m33s)  kubelet            Container image "quay.io/keycloak/keycloak:17.0.0" already present on machine
...
  Warning  BackOff    2m8s (x13 over 3m31s)  kubelet            Back-off restarting failed container
$ oc logs keycloak-1-6pmv9
Keycloak - Open Source Identity and Access Management

Find more information at: https://www.keycloak.org/docs/latest

Usage:

kc.sh [OPTIONS] [COMMAND]
...
  Please, take a look at the documentation for more details before deploying in
production.

Use "kc.sh start --help" for the available options when starting the server.
Use "kc.sh <command> --help" for more information about other commands.
```
Looks like quay.io/keycloak/keycloak:17.0.0 does not set additional necessary arguments for kc.sh to start up. It only sets:
```
      Entrypoint:
      - /opt/keycloak/bin/kc.sh
      Env:
      - PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
      - container=oci
```
Either quay.io/keycloak/keycloak:17.0.0 should be updated with full args for start-up, or openshift-examples/keycloak.yaml needs be updated with those full args.

### Version

17.0.0

### Expected behavior

See above, pod should be Running

### Actual behavior

See above, pod is CrashLoopBackOff

### How to Reproduce?

See above

### Anything else?
None

DGuhr:
Draft PR as long as I haven't found the changes to the actual guide ;)

Closes #300

### DIFF
--- a/openshift-examples/keycloak.yaml
+++ b/openshift-examples/keycloak.yaml
@@ -4,24 +4,24 @@ metadata:
   name: keycloak
   namespace: openshift-config
   annotations:
-    description: An example Keycloak server with HTTPS
+    description: An example template for trying out Keycloak on OpenShift
     iconClass: icon-sso
     openshift.io/display-name: Keycloak
     tags: keycloak
-    version: 9.0.4-SNAPSHOT
+    version: 17.0.0-SNAPSHOT
 objects:
   - apiVersion: v1
     kind: Service
     metadata:
       annotations:
-        description: The web server's https port.
+        description: The web server's http port.
       labels:
         application: '${APPLICATION_NAME}'
       name: '${APPLICATION_NAME}'
     spec:
       ports:
-        - port: 8443
-          targetPort: 8443
+        - port: 8080
+          targetPort: 8080
       selector:
         deploymentConfig: '${APPLICATION_NAME}'
   - apiVersion: v1
@@ -29,14 +29,14 @@ objects:
     kind: Route
     metadata:
       annotations:
-        description: Route for application's https service.
+        description: Route for application's service.
       labels:
         application: '${APPLICATION_NAME}'
       name: '${APPLICATION_NAME}'
     spec:
       host: '${HOSTNAME_HTTPS}'
       tls:
-        termination: passthrough
+        termination: edge
       to:
         name: '${APPLICATION_NAME}'
   - apiVersion: v1
@@ -60,21 +60,17 @@ objects:
         spec:
           containers:
             - env:
-                - name: KEYCLOAK_USER
-                  value: '${KEYCLOAK_USER}'
-                - name: KEYCLOAK_PASSWORD
-                  value: '${KEYCLOAK_PASSWORD}'
-                - name: DB_VENDOR
-                  value: '${DB_VENDOR}'
-                - name: JGROUPS_DISCOVERY_PROTOCOL
-                  value: dns.DNS_PING
-                - name: JGROUPS_DISCOVERY_PROPERTIES
-                  value: 'dns_query=${APPLICATION_NAME}.${NAMESPACE}.svc.cluster.local'
+                - name: KEYCLOAK_ADMIN
+                  value: '${KEYCLOAK_ADMIN}'
+                - name: KEYCLOAK_ADMIN_PASSWORD
+                  value: '${KEYCLOAK_ADMIN_PASSWORD}'
+                - name: KC_PROXY
+                  value: '${KC_PROXY}'
               image: quay.io/keycloak/keycloak:17.0.0
               livenessProbe:
                 failureThreshold: 3
                 httpGet:
-                  path: /auth/realms/master
+                  path: /realms/master
                   port: 8080
                   scheme: HTTP
                 initialDelaySeconds: 60
@@ -82,18 +78,16 @@ objects:
               ports:
                 - containerPort: 8080
                   protocol: TCP
-                - containerPort: 8443
-                  name: https
-                  protocol: TCP
               readinessProbe:
                 failureThreshold: 10
                 httpGet:
-                  path: /auth/realms/master
+                  path: /realms/master
                   port: 8080
                   scheme: HTTP
                 initialDelaySeconds: 30
               securityContext:
                 privileged: false
+              args: ["start-dev"]
       triggers:
         - type: ConfigChange
 parameters:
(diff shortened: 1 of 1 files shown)


openshift-examples/keycloak.yaml
```diff
@@ -4,39 +4,39 @@ metadata:
...
+        description: The web server's http port.
       labels:
         application: '${APPLICATION_NAME}'
       name: '${APPLICATION_NAME}'
     spec:
       ports:
-        - port: 8443
-          targetPort: 8443
+        - port: 8080
+          targetPort: 8080
```
xingxingxia, line 24: I don't think I agree. It is always to give secure examples for use. Can we update by keeping secure port working?

DGuhr: Please keep in mind these are Quick Starts, not a production guide. We're using `edge` in other examples too, and align this example now to work like the others. see e.g. https://raw.githubusercontent.com/keycloak/keycloak-quickstarts/latest/kubernetes-examples/keycloak.yaml 

For making kc ready for production, we have https://www.keycloak.org/server/configuration-production now.

openshift-examples/keycloak.yaml
```diff
@@ -60,40 +60,34 @@ objects:
         spec:
           containers:
             - env:
-                - name: KEYCLOAK_USER
-                  value: '${KEYCLOAK_USER}'
-                - name: KEYCLOAK_PASSWORD
-                  value: '${KEYCLOAK_PASSWORD}'
-                - name: DB_VENDOR
-                  value: '${DB_VENDOR}'
-                - name: JGROUPS_DISCOVERY_PROTOCOL
```
andreaTP, line 69: is it intentional to drop the clustering configuration?

In case you wanna preserve it, the new variables are:
```yaml
- name: jgroups.dns.query
  value: 'dns_query=${APPLICATION_NAME}.${NAMESPACE}.svc.cluster.local'
- name: KC_CACHE
  value: 'ispn'
- name: KC_CACHE_STACK
  value: 'kubernetes'
```
cc. @pedroigor

andreaTP (changes requested):
I have a note on the caching configuration, the rest LGTM

openshift-examples/keycloak.yaml
vmuzikar: Not sure if this was even working before. I don't see discovery service defined here e.g.

We're also not exposing number of replicas as a param, so we technically don't support clustering. We could of course add it but it's out of scope of this PR.

openshift-examples/keycloak.yaml
```diff
@@ -102,22 +96,22 @@ parameters:
...
     from: '[a-zA-Z0-9]{8}'
     required: true
-  - name: DB_VENDOR
-    displayName: DB Vendor
-    description: 'DB vendor (H2, POSTGRES, MYSQL or MARIADB)'
-    value: H2
+  - name: KC_PROXY
+    displayName: Keycloak proxy configuration
+    description: 'Keycloak Proxy configuration (edge, passthrough, reencrypt, none)'
+    value: edge
```
vmuzikar, line 114: Does it actually make really sense to expose this? Anything but `edge` requires additional config (TLS keys mounting) which is not exposed, is that correct? If so I'd say we might need to expose it all for the proxy to be meaningful.

Alternative approach might be hardcoding a TLS Secret that is generated by OpenShift itself, similarly how it's implemented in the [RH-SSO templates](https://github.com/jboss-container-images/redhat-sso-7-openshift-image/blob/861625fa0aa1ad4c013174223566fa2b4bd34090/templates/sso75-ocp4-x509-https.json#L152).

In any case, SSL config is out of scope of this. But without SSL config, proxy config makes only little sense to me. So I'd suggest removing the proxy param and addressing it as a follow-up.

openshift-examples/keycloak.yaml
```diff
@@ -102,22 +96,22 @@ parameters:
...
     required: true
   - name: HOSTNAME_HTTPS
```
vmuzikar, line 116: Since we're not using TLS, please rename this param.

vmuzikar (changes requested):
Nice work, @DGuhr.

Have you tested it with an OCP instance? If not, I can help.

@iankko I know your super busy, but just in case you found a few minutes to spare, can you please give it a look too? You're most familiar with OCP Templates.

openshift-examples/keycloak.yaml
```diff
@@ -60,40 +60,34 @@ objects:
         spec:
           containers:
             - env:
-                - name: KEYCLOAK_USER
-                  value: '${KEYCLOAK_USER}'
-                - name: KEYCLOAK_PASSWORD
-                  value: '${KEYCLOAK_PASSWORD}'
-                - name: DB_VENDOR
-                  value: '${DB_VENDOR}'
```
vmuzikar, line 68: Also, how is the DB supposed to be configured now? Should we add some params to the template?

Before, it started OOTB thanks to H2. This is not the case, right?

openshift-examples/keycloak.yaml
```diff
@@ -60,40 +60,34 @@ objects:
...
-                - name: JGROUPS_DISCOVERY_PROTOCOL
-                  value: dns.DNS_PING
-                - name: JGROUPS_DISCOVERY_PROPERTIES
-                  value: 'dns_query=${APPLICATION_NAME}.${NAMESPACE}.svc.cluster.local'
+                - name: KEYCLOAK_ADMIN
+                  value: '${KEYCLOAK_ADMIN}'
+                - name: KEYCLOAK_ADMIN_PASSWORD
+                  value: '${KEYCLOAK_ADMIN_PASSWORD}'
+                - name: KC_PROXY
+                  value: '${KC_PROXY}'
```
andreaTP, line 68: hardcode the param to `edge`

andreaTP (commented):
test everything on OpenShift

vmuzikar:
Replaced by #306.

PR #306, opened 2022-03-16
Fix keycloak.yaml

### LINKED ISSUE #300: Latest openshift-examples/keycloak.yaml of 17.0.0 does not work

andreaTP:
Overtakes #304 

Closes #300

Co-authored-by: Dominik Guhr <dguhr@redhat.com>
Co-authored-by: Andrea Peruffo <andrea.peruffo1982@gmail.com>

### DIFF
--- a/openshift-examples/keycloak.yaml
+++ b/openshift-examples/keycloak.yaml
@@ -2,26 +2,25 @@ kind: Template
 apiVersion: template.openshift.io/v1
 metadata:
   name: keycloak
-  namespace: openshift-config
   annotations:
-    description: An example Keycloak server with HTTPS
+    description: An example template for trying out Keycloak on OpenShift
     iconClass: icon-sso
     openshift.io/display-name: Keycloak
     tags: keycloak
-    version: 9.0.4-SNAPSHOT
+    version: 17.0.0-SNAPSHOT
 objects:
   - apiVersion: v1
     kind: Service
     metadata:
       annotations:
-        description: The web server's https port.
+        description: The web server's http port.
       labels:
         application: '${APPLICATION_NAME}'
       name: '${APPLICATION_NAME}'
     spec:
       ports:
-        - port: 8443
-          targetPort: 8443
+        - port: 8080
+          targetPort: 8080
       selector:
         deploymentConfig: '${APPLICATION_NAME}'
   - apiVersion: v1
@@ -29,14 +28,14 @@ objects:
     kind: Route
     metadata:
       annotations:
-        description: Route for application's https service.
+        description: Route for application's service.
       labels:
         application: '${APPLICATION_NAME}'
       name: '${APPLICATION_NAME}'
     spec:
-      host: '${HOSTNAME_HTTPS}'
+      host: '${HOSTNAME}'
       tls:
-        termination: passthrough
+        termination: edge
       to:
         name: '${APPLICATION_NAME}'
   - apiVersion: v1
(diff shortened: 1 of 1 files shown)


openshift-examples/keycloak.yaml
```diff
@@ -60,40 +60,40 @@ objects:
...
                 initialDelaySeconds: 30
               securityContext:
                 privileged: false
+              volumeMounts:
+                - mountPath: /opt/keycloak/data
+                  name: empty
+              args: ["start-dev"]
+          volumes:
+            - name: empty
+              emptyDir: {}
```
vmuzikar, line 96: jFTR, created #307.

vmuzikar (approved):

openshift-examples/keycloak.yaml
```diff
@@ -4,39 +4,39 @@ metadata:
   name: keycloak
   namespace: openshift-config
```
iankko, line 5: @andreaTP Maybe OT / unrelated to fixing the original issue, but does the namespace field needs to be present there?

iankko (dismissed):
LGTM AFAICT Confirmed it works on both OCP v4.9 & also on OCP v3.11 (after removing / commenting out the `namespace:` field on line #5).

@andreaTP 
Just small suggestion / nitpick, if you remove the `namespace: openshift-config` line / field, the `oc create -f keycloak.yaml` will create the template in the current namespace, than to create it in the `openshift-config` each time. The template being created in the current namespace is IMHO a better UX experience, since it can be used immediately without needing to specify namespaces in subsequent commands


If you wanted to specify a namespace, where it should be created, you can always specify `-n` parameter to the `oc create -f` command, instead of using a hardcoded / fixed namespace namespace in the template itself.

openshift-examples/keycloak.yaml
andreaTP: removed 👍

vmuzikar (changes requested):
I missed this before. This needs to go to the `main` branch, not `latest`. See https://github.com/keycloak/keycloak-quickstarts/blob/latest/CONTRIBUTING.md#picking-correct-branch-for-your-pr

andreaTP:
@vmuzikar fixed

openshift-examples/keycloak.yaml
vmuzikar: @andreaTP With https://github.com/keycloak/keycloak/pull/10783 merged, I think we can remove the volume mount. WDYT?

vmuzikar (changes requested):

openshift-examples/keycloak.yaml
andreaTP: @vmuzikar we need a release before removing this.

vmuzikar: This is `main` branch, not `latest`. That means this PR is already targeted for the next release. 🙂

andreaTP: The image used in `keycloak.yaml` is `17.0.0`, we can possibly remove the workaround when we update the image version.

vmuzikar: Again, `main` branch is the dev branch. 🙂 It's a preparation for 18.0.0 regardless to what's currently in the `image` field. So it should be ok to change it now. Is that correct @stianst?

DGuhr (approved):

openshift-examples/keycloak.yaml
andreaTP: I think this is a not important detail:
 - mounting an external volume is actually good practice
 - when we switch to the next version in the image field we can remove the workaround (otherwise we are going to have a broken template again for the time being)
 
 Overall, whatever works for you.

vmuzikar: Agree it's not a very important detail. It just feels a bit wrong to be introducing a "workaround" for a problem of the past. 😉

It is ok to break the template in `main` branch as it is a "future" branch for the next release. That's a standard process. Users are supposed to use `latest` branch.

But ok, I don't want to block this PR, so approving.

vmuzikar (approved):

abstractj (approved):

PR #316, opened 2022-04-22
Added instructions and support for have script policy provider workin…

mposolda:
…g on Java 17

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,5 +61,4 @@ target
 node_modules
 
 # Keycloak distribution
-keycloak
 keycloak-dist
--- a/app-authz-photoz/README.md
+++ b/app-authz-photoz/README.md
@@ -116,6 +116,30 @@ Click on the `Select file` button, which means you want to import a resource ser
     
 Now click `Upload` and the resource server will be updated accordingly.
 
+Deploy the script engine on Java 15 or later
+-----------
+In case your Keycloak server is running on Java 15 or later, you need to install script engine to your Keycloak server.
+Otherwise JavaScript policies won't work.
+
+If you use Keycloak server running on Wildfly, you can install script engine by adding new module nashorn-core to your Keycloak server.
+When server is up and running, you can run commands similar to these in the KEYCLOAK_HOME/bin directory:
+
+```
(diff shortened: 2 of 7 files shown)


mposolda:
This is draft PR related to https://github.com/keycloak/keycloak/pull/11322 . See that PR for more details.

authz-js-policies/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+What is it?
+-----------
+
+TODO
+
+System Requirements
+-------------------
+
+You need to have Keycloak 18.0.Final running.
```
DGuhr, line 19: perhaps better "at least Keycloak 18.0.Final" or sth? I'd avoid strict version pinning in a general example.

authz-js-policies/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+All you need to build this project is Java 8.0 (Java SDK 1.8) or later and Maven 3.3.3 or later.
```
DGuhr, line 21: For Quarkus, it'd be Java11

authz-js-policies/README.md
mposolda: Yes, true. Thanks for these points. I've copy/pasted this from the README of some other example :-) Similarly the Java version.

It seems that README files of the quickstarts, which are deployable to Keycloak server, may need some revision as they are currently target just for the Keycloak on Wildfly.

This particular quickstart didn't have README file like the other quickstarts have. So I've added README file, but for this README, I've added lots of TODO and I focused only on parts specific to instructions for deployment on Keycloak on Wildfly VS Keycloak on Quarkus as this is the most important part for the particular use-case addressed here for the scripts providers. I can properly update the README file and remove all TODOs if we agree that my approach for script providers work.

pedroigor (approved):

mposolda:
@pedroigor @DGuhr Do you please have a chance to re-review this? Corresponding main codebase PR is also ready for the review: https://github.com/keycloak/keycloak/pull/11322 .
Thanks

pedroigor (approved):

PR #323, opened 2022-05-23
Extracting legacy store module (and refactor credential management)

ahus1:
This PR accompanies PR keycloak/keycloak#10279


ahus1:
Method added here will be added to the base class; no changes necessary to make the existing example compile with the new code changes. Closing this PR.

PR #328, opened 2022-06-09
Fix compile issues with event-store-mem example

### LINKED ISSUE #327: On latest, the module event-store-mem doesn't compile any more
### Describe the bug

The API change has been introduced in this issue: 
https://github.com/keycloak/keycloak/issues/11189

### Version

999-SNAPSHOT

ahus1:
Closes #327

### DIFF
--- a/event-store-mem/src/main/java/org/keycloak/quickstart/event/storage/MemEventStoreProvider.java
+++ b/event-store-mem/src/main/java/org/keycloak/quickstart/event/storage/MemEventStoreProvider.java
@@ -25,6 +25,7 @@ import org.keycloak.events.admin.AdminEvent;
 import org.keycloak.events.admin.AdminEventQuery;
 import org.keycloak.events.admin.OperationType;
 import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.RealmModel;
 
 import java.util.Iterator;
 import java.util.LinkedList;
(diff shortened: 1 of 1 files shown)


ahus1:
With this change in place, `event-store-mem` is no longer on the list of failed examples. 

Other examples still fail, and will be fixed in other PRs.

@mhajas - as you've been involved in the original PR that included that API change, I ask you to review this PR. Thanks!

vmuzikar (approved):
@ahus1 LGTM. Thank you for your effort to make QS CI green again! 🙂

vmuzikar:
One thing I was wondering about. With this QS broken, does it mean we did some breaking changes to our SPI? Was that intended?

ahus1:
Hi @vmuzikar - at first sight this is a private SPI, so we're ok to change it. 

I asked @mhajas for a review here, he might add additional insights.

mhajas (approved):
> > One thing I was wondering about. With this QS broken, does it mean we did some breaking changes to our SPI? Was that intended?

> Hi @vmuzikar - at first sight this is a private SPI, so we're ok to change it.

Just a small addition, there are two interfaces for events - `EventListenerProvider` and `EventStoreProvider`. The first one is to listen to events, this is implemented when somebody wants to listen to events. The latter is just for events management in the Admin console. What changed are only methods from the latter, hence we most probably didn't break any existing community code because I would say people are more interested in listening to the events than replacing the store that is used for displaying the events in the Admin console.

ahus1:
@mhajas - no worries. Getting this merged is one step to get the job back in the green, and then we'll no longer miss anything that might break.

PR #340, opened 2022-06-22
Reduce number of warnings in Maven build to avoid distractions

### LINKED ISSUE #339: Reduce the number of Maven related warnings in the build
### Describe the bug

The following messages appear in the log, and were a distraction when I was fixing other problems with this build. That's why I raise this issue.

Those warning were present multiple times in the output when building the project:

```
Warning:  Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
Warning:  File encoding has not been set, using platform encoding UTF-8, i.e. build is platform dependent!
Warning:  Rule 0: org.apache.maven.plugins.enforcer.RequireFilesExist failed with message:
Some required files are missing:
/home/runner/work/keycloak-quickstarts/keycloak-quickstarts/app-authz-photoz/photoz-restful-api/src/test/resources/quickstart-realm.json
Warning:  JAR will be empty - no content was marked for inclusion!
```

### Version

999-SNAPSHOT

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

ahus1:
Closes #339

### DIFF
--- a/app-authz-photoz/photoz-testsuite/pom.xml
+++ b/app-authz-photoz/photoz-testsuite/pom.xml
@@ -25,7 +25,7 @@
         <relativePath>../pom.xml</relativePath>
     </parent>
     <artifactId>photoz-testsuite</artifactId>
-    <packaging>jar</packaging>
+    <packaging>pom</packaging>
     <name>Keycloak Quickstarts: Photoz Testsuite</name>
     <description>Photoz Testsuite</description>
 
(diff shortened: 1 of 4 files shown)


sguilhen (approved):
LGTM, thank you @ahus1

app-authz-photoz/photoz-testsuite/pom.xml
```diff
@@ -25,7 +25,7 @@
         <relativePath>../pom.xml</relativePath>
     </parent>
     <artifactId>photoz-testsuite</artifactId>
-    <packaging>jar</packaging>
+    <packaging>pom</packaging>
```
hmlnarik, line 28: Out of curiosity - why is this change here?

ahus1: This is a project with only tests, and having it packaged as a JAR will make Maven complain with "Warning:  JAR will be empty - no content was marked for inclusion!"

So marking it as a POM will get rid of that warning.

PR #343, opened 2022-06-24
Log information from the keycloak.log, so they are available in the GitHub action output

### LINKED ISSUE #342: Add information about failed Keycloak start
### Description

When the Keycloak startup fails, the only log provided is the following: 

```
Starting Keycloak.................................................. timeout!
Error: Process completed with exit code 1.
```

This is not helpful to find the cause. Therefore it should output the log of the Keycloak server that might show exceptions or other information.

### Discussion

main

### Motivation

Provide more information to analyze build failures.

### Details

_No response_

ahus1:
Closes #342

### DIFF
--- a/scripts/start-local-server.sh
+++ b/scripts/start-local-server.sh
@@ -18,6 +18,7 @@ function waitForServer {
       sleep 1
     else
       echo " timeout!"
+      cat keycloak.log
       exit 1
     fi
   done


ahus1:
The log will only contain the startup messages of the Keycloak instance. When this has been started and startup failed, all messages would be important to see. No test will have run there, yet, so there won't be much logs in the file. A local startup file has ~150 lines, and 24 k.

kami619:
the change looks good to me, to log info in case of an abnormal server start up

martin-kanis (approved):

PR #345, opened 2022-06-29
Change enum ResourceType to interface with String constants

vramik:
Relates to https://github.com/keycloak/keycloak/pull/12698
Fixes https://github.com/keycloak/keycloak/issues/12485


hmlnarik (approved):
Thank you @vramik !

PR #355, opened 2022-08-30
Update QuickStarts test to use Quarkus distribution, and nightly builds

stianst:


stianst:
First pass at migrating the QuickStarts to Quarkus. I'd like to get this one merged quickly, then have a number of follow-ups:

* Remove need to add '/auth'
* Update docs
* Fix/migrate the the tests that are commented out in run-tests.sh by this PR
* Have a better way to check when server is up and running than simply sleeping for 10 seconds ;)

At least with this PR we have most of the tests passing with the Quarkus dist.

stianst:
I've already updated the workflow to run on a schedule after the nightly builds have been released, which will allow us to continuously monitor if the QuickStarts break

pedroigor (approved):

stianst:
Follow-up issues:
https://github.com/keycloak/keycloak-quickstarts/issues/357
https://github.com/keycloak/keycloak-quickstarts/issues/356

PR #363, opened 2022-10-23
ReadMe-Changes

Toma5OD:
Additions made to the Read-Me files due to UI changes implemented with Keycloak 19.x

File added into config for Client import with Bearer Only set to True.
Bearer-Only option removed with UI change from Keycloak 18.x to 19.x


Toma5OD:
Not required due to transition from Java-EE to Jakarta

PR #396, opened 2023-04-12
Issue 19540 quickstart

pedroigor:


app-authz-jakarta-servlet/config/realm-import.json
```diff
@@ -82,11 +80,11 @@
...
       "clientId": "authz-servlet",
       "enabled": true,
-      "baseUrl": "http://localhost:8080/authz-servlet",
-      "adminUrl": "http://localhost:8080/authz-servlet",
+      "baseUrl": "http://127.0.0.1:8080/authz-servlet",
+      "adminUrl": "http://127.0.0.1:8080/authz-servlet",
       "bearerOnly": false,
       "redirectUris": [
-        "http://localhost:8080/authz-servlet/*"
+        "*"
```
mposolda, line 87: I wonder why is this needed? We mention in the "Threat Docs" https://www.keycloak.org/docs/latest/server_admin/index.html#open-redirectors to avoid wildcards, so does not look very great to use them in the quickstart?

Also is it strictly needed to change `baseUrl` and `adminUrl` to `127.0.01` instead of `localhost`? Is it just because of the tests or any other reason for this?

app-authz-jakarta-servlet/pom.xml
```diff
@@ -35,55 +35,69 @@
...
+            <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.apache.httpcomponents</groupId>
-            <artifactId>httpclient</artifactId>
-            <version>4.5</version>
-            <scope>compile</scope>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-admin-client-jakarta</artifactId>
+            <version>21.0.2</version>
```
mposolda, line 65: I guess the hardcoded version here should be avoided?

Is it possible to use `<version>${version.keycloak}</version>` ? Or ideally avoid `<version>` element entirely and make sure that correct version is just taken from the parent pom (keycloak BOM) as I've mentioned in the other comments to this file?

app-authz-jakarta-servlet/pom.xml
```diff
@@ -35,55 +35,69 @@
...
+            <artifactId>jakarta.servlet-api</artifactId>
+            <version>6.0.0</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.keycloak</groupId>
-            <artifactId>keycloak-adapter-core</artifactId>
-            <scope>provided</scope>
+            <artifactId>keycloak-policy-enforcer</artifactId>
+            <version>${version.keycloak}</version>
```
mposolda, line 52: Minor: Do we need this `<version>` element? Could it just take it automatically from parent pom (imported from keycloak bom)?

app-authz-jakarta-servlet/pom.xml
```diff
@@ -35,55 +35,69 @@
...
         </dependency>
+
+        <!-- Test Dependencies -->
         <dependency>
             <groupId>org.keycloak</groupId>
-            <artifactId>keycloak-adapter-spi</artifactId>
-            <scope>provided</scope>
+            <artifactId>keycloak-client-registration-api</artifactId>
+            <version>${version.keycloak}</version>
```
mposolda, line 59: Minor: Same like above

app-authz-jakarta-servlet/src/test/resources/quickstart-realm.json
```diff
@@ -86,7 +84,7 @@
       "adminUrl": "http://127.0.0.1:8080/authz-servlet",
       "bearerOnly": false,
       "redirectUris": [
-        "http://127.0.0.1:8080/authz-servlet/*"
+        "*"
```
mposolda, line 87: Same comment like above for the other file

mposolda (changes requested):
@pedroigor Added few inline comments. Marking this as "request changes" until we figure how to tackle this.

app-authz-jakarta-servlet/config/realm-import.json
```diff
@@ -82,11 +80,11 @@
     {
       "clientId": "authz-servlet",
       "enabled": true,
-      "baseUrl": "http://localhost:8080/authz-servlet",
-      "adminUrl": "http://localhost:8080/authz-servlet",
+      "baseUrl": "http://127.0.0.1:8080/authz-servlet",
```
mposolda, line 83: Any specific reason why all those URLs need to be changed from `localhost` to `127.0.01` ?

app-authz-jakarta-servlet/src/test/resources/quickstart-realm.json
mposolda: I am unresolving this comment as wildcard is still here? Any reason why it is needed and was not needed before with old adapter?

mposolda (changes requested):
@pedroigor Still few inline questions added

app-authz-jakarta-servlet/src/test/resources/quickstart-realm.json
pedroigor: Hope is fine now.

app-authz-jakarta-servlet/config/realm-import.json
pedroigor: Hope is fine now.

pedroigor:
Hope is fine now.

mposolda (approved):
@pedroigor Thanks for the updates!

PR #400, opened 2023-04-18
Fix listing of users in user-storage-simple

### LINKED ISSUE #399: Fix listing of users in user-storage-simple
### Describe the bug

In user-storage-simple there is not possible to list all users (including those from the federated provider). When searching by * only local users are returned. When searching by username (malcom) the user is found.

### Version

999.0.0-Snapshot

### Expected behavior

It should be possible to list all users even when only searching by username is supported.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

martin-kanis:
Closes #399


vramik (approved):
Thanks @martin-kanis

mhajas (approved):
Approving per @vramik 's approval

PR #403, opened 2023-04-20
Removing redundant quickstarts

pedroigor:
Removing the following quickstarts:
 
* app-authz-jee-vanilla
* app-authz-photoz
* app-authz-springboot
* app-authz-uma-photoz
* app-jee-html5
* app-jee-jsp
* app-profile-jee-jsp
* app-profile-jee-html5
* app-profile-jee-vanilla

Those related to JEE/Servlet/JSP should be covered by `app-profile-jee-jsp`.

Those related to UMA are being removed as we are not sure about the roadmap for UMA support and we should start discouraging people from using it.

Those related to Spring and Authz should be reworked as part of Spring and Authz support.

Those related to HTML5 are already covered by quickstarts using Node.JS.


mposolda:
@pedroigor LGTM. Few questions: 

How about:
- Some more spring stuff: `app-authz-springboot-multitenancy`, `app-springboot`
- Some more JEE stuff: `app-profile-saml-jee-jsp`, `service-jee-jaxrs`

Do you prefer to keep those for now?

pedroigor:
I'm also removing:

* `app-authz-springboot-multitenancy`
* `app-springboot`
* `service-jee-jaxrs`

Not removing:

* `app-profile-saml-jee-jsp`

For SAML, we want a quickstart but it should be migrated to Jakarta soon.

mposolda (approved):

mposolda:
> I'm also removing:
> 
>     * `app-springboot`
> 
>     * `app-authz-springboot-multitenancy`
> 
> 
> Not removing:
> 
>     * `app-profile-saml-jee-jsp`, we need SAML examples but migrate to Jakarta so that we run with latest Wildfly/EAP
> 
>     * `service-jee-jaxrs`, this one is used by other quickstarts and we need to review it in the next iteration. If we remove it now, we also need to change other quickstarts.
> 
> 
> Works for you?

@pedroigor Yes, Thanks for the pointers. I don't have anything else and I am approving now.

stianst (approved):

stianst:
LGTM as a follow-up at some point we can consider making `app-nodejs-html5` runnable in different ways. It's just a bunch of static web pages, so could be deployed to any web server, not just using Node.js.

pedroigor:
@stianst For now, I've created a `js` category to our quickstarts so that you can infer it is pure JS (although the quickstart uses Node.js to deploy the static resources)

Issue #402, opened 2023-04-19
Remove hardcoded version on keycloak-admin-client-jakarta from app-authz-jakarta-servlet/pom.xml


mposolda:
### Description

This is follow-up on https://github.com/keycloak/keycloak-quickstarts/pull/396 .

There is hardcoded of `keycloak-admin-client-jakarta` to `21.0.2` in `app-authz-jakarta-servlet/pom.xml` (and also in `app-authz-rest-springboot/pom.xml` after the PR https://github.com/keycloak/keycloak-quickstarts/pull/405/files is merged). This issue is about to replace that hardcoded versions with the same version as other Keycloak dependencies in quickstarts. The version should be ideally taken from Keycloak bom or somewhere...

mposolda:
Closed by commit https://github.com/keycloak/keycloak-quickstarts/commit/d563da6ef0cc0dde2b46e03426a5eee0063fa18a

PR #405, opened 2023-04-27
Example about how to use the policy enforcer with Spring

pedroigor:
* Example about how to integrate the PEP in Spring Boot/Security applications
* Removing `app-authz-rest-employee` because it shows the same thing as the new example
* Temporarily enabling GHA for Spring quickstarts as they require the repository using JDK 17

This change set requires https://github.com/keycloak/keycloak/pull/20003.


app-authz-rest-springboot/pom.xml
```diff
@@ -1,128 +1,92 @@
...
             <groupId>org.keycloak</groupId>
-            <artifactId>keycloak-test-helper</artifactId>
-            <version>${version.keycloak}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>net.sourceforge.htmlunit</groupId>
-            <artifactId>htmlunit</artifactId>
+            <artifactId>keycloak-admin-client-jakarta</artifactId>
+            <version>21.0.2</version>
```
mposolda, line 78: Same like before, but I guess it is fine for this PR as we discussed? I've added comment to https://github.com/keycloak/keycloak-quickstarts/issues/402

mposolda (approved):
@pedroigor I am approving, Thanks for the updates!

I've added one minor comment about hardcoded version, but that is not a blocker for this PR as we already have follow-up issue.

Another thing are failing test. I guess we may need to wait for the latest nightly release until https://github.com/keycloak/keycloak/commit/79cd47a28030cde393f8093bee67cd5867099df8 is available in the Keycloak nightly? Other alternative is to just merge this PR as tests should be "autofixed" later once stuff is in Keycloak nightly. My preference is to wait, but leaving to you (not sure if this PR blocks you from other things?).

app-authz-rest-springboot/pom.xml
pedroigor: Yeah, both this one and the error are related to how this repo is caching artifacts when executing the workflow.

PR #406, opened 2023-04-28
Added admin-ui customization 

agagancarczyk:
This is related to https://github.com/keycloak/keycloak/issues/19969


extend-admin-ui-quick-start/README.md
```diff
@@ -0,0 +1,57 @@
...
+├── mytheme/
+│ ├── admin/
+│ │ ├── resources/
+│ │ | ├── assets/
+│ │ | ├── css/
+│ │ | | ├── styles.css
+│ │ | ├── resources/
+│ │ ├── theme.properties
+
+In order to add a custom theme, in the `js/apps/keycloak-server/server/themes` folder, create a new folder named `mytheme` (can be any name you desire) which houses your personalized Admin UI theme. Within the `admin` sub-folder of this `mytheme` folder, there is a `theme.properties` file, along with a `resources` sub-folder that encompasses directories for CSS files, JavaScript files, and images. Within the `css` sub-folder, there is a `styles.css` file, which allows the addition of customized PatternFly styles. The `theme.properties` is a configuration file defines attributes such as the parent theme, and theme css. The `theme.properties` file is used to register the theme with the Keycloak server and integrate it with the Keycloak user interface.
```
edewit, line 22: maybe we don't want to use `js/apps/keycloak-server/server/themes` here but instead just tell people to put it into `themes` folder of the keycloak server

extend-admin-ui-quick-start/README.md
```diff
@@ -0,0 +1,57 @@
...
+5. Create a custom stylesheet `(styles.css)` in the `resources/css` folder to contain your custom styles and overrides.
+6. Define your custom `styles=css/styles.css` in the `theme.properties` file. Make sure to add `parent=keycloak.v2` as parent theme for this custom theme. `title` of the Admin ui can be changed as well as `logo`.
+
+By following these best practices, you can ensure that your custom Keycloak theme remains consistent with the Keycloak Admin UI and is easy to maintain and update.
+
+### Applying a Customized Theme
+
+Start the Keycloak server by running the command `npm run start --workspace=keycloak-server`. To launch the Admin UI run the command `npm run dev`. After that, log in by accessing `http://localhost:8080/` and navigate to the `Themes` tab located in `Realm settings`. From there, change the Admin UI theme to your preferred custom theme. Finally, navigate to `http://localhost:8180/` to view the style you created being applied.
```
edewit, line 41: Here I would also recommend just to tell to start the server and not this specific ui development commands. It's not necessary to run `npm run dev` and the server can be started with `./kc.sh` or `./kc.bat`

edewit (changes requested):
maybe there is also a way to not have the entire build of the admin ui in the assets folder

extend-admin-ui-quick-start/README.md
ssilvert: > maybe we don't want to use `js/apps/keycloak-server/server/themes` here but instead just tell people to put it into `themes` folder of the keycloak server
+1  We definitely do not want to pollute `js/apps/keycloak-server/server/themes`.

A couple of things to keep in mind here:
1. For final deployment, you should have the option to bundle as a jar.  We don't need to have that ability in this PR, but the quickstart should be able to build the theme as a jar.  See https://www.keycloak.org/docs/latest/server_development/#deploying-themes
2. In the WildFly version of Keycloak, it was possible to have more than one /themes directory.  So it would be possible to tell your server to point to your development directory when creating a theme.  I'm not sure if the Quarkus version has this, but if it doesn't we can add it if needed.

agagancarczyk: Thanks @ssilvert I will create the next PR with a suggested solution of adding an option to download a Keycloak server and provide a way to start it with the --spi-theme-folder-dir=<path> and using Vite.

extend-admin-ui-quick-start/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+![Keycloak Clients Page](./img/keycloakClients.png "Keycloak Clients Page")
+
+![Keycloak Create A New Client Page](./img/keycloakNewClient.png "Keycloak Create A New Client Page")
+
+**Built-in themes should not be modified directly, instead a custom theme should be created.**
+
+## Disclaimer
+
+The way PatternFly variables are used in the Admin UI is subject to change, as we may update and modify the design system to better suit the needs of our users. Additionally, PatternFly, the design system on which our UI is based, may also make changes to their variables in future updates. As such, any information, analysis, or conclusions drawn from the current use of PatternFly variables in the Admin UI may not necessarily hold true or remain valid in the future. It is important to consider the potential for changes in the way we use PatternFly variables and any updates to the PatternFly design system, and to update any relevant information accordingly.
```
ssilvert, line 44: I'd like to get rid of this disclaimer.  More importantly, we should be vigilant about the need for the disclaimer.  

One of our primary goals is to allow customizations that do not break between releases.  Obviously, we can't make hard promises about this, especially when it comes to PatternFly.  But I'd rather say that we will make every effort to ensure forward compatibility then make it sound like we expect to fail in our goal.

For now, let's just remove the disclaimer.  We can state our positive intentions later.

extend-admin-ui-quick-start/themes/theme/admin/theme.properties
```diff
@@ -0,0 +1,11 @@
+# This theme will inherit everything from its parent unless
+# it is overridden in the current theme.
+parent=keycloak.v2
+
+# This is the title of the Admin UI.
+# It can be any title.
+title=My Admin UI Title
+
+# Look at the styles.css file to see examples of using PatternFly's CSS variables
+# for modifying look and feel.
```
ssilvert, line 10: Note that multiple stylesheets can be listed using a space as a delimiter.

ssilvert (changes requested):

extend-admin-ui-quick-start/README.md
```diff
@@ -0,0 +1,44 @@
...
+### Required Folder Structure
+
+The recommended folder structure for a custom Keycloak theme is as follows:
+
+├── theme/
+│ ├── admin/
+│ │ ├── resources/
+│ │ | ├── css/
+│ │ | | ├── styles.css
+│ │ ├── theme.properties
```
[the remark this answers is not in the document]
agagancarczyk: Thanks @ssilvert I can remove the Readme and update the official Keycloak documentation instead. Wdyt?

extend-admin-ui-quick-start/README.md
agagancarczyk: Thanks @ssilvert , I agree. I'll remove this for now and we can add something later.

extend-admin-ui-quick-start/README.md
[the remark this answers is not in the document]
agagancarczyk: @ssilvert I have just pushed an updated. Please double-check if this is ok now. Thanks!

extend-admin-console-quick-start/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Creating Themes
+
+With themes, it is possible to configure the Admin Console's look and feel. To ensure consistency and maintainability of the custom Keycloak theme, we advise using PatternFly best practices when customizing the Keycloak Admin Console. This includes following the recommended folder structure and using PatternFly CSS classes and guidelines when creating custom styles and overrides.
+
+### Required Folder Structure
+
+The recommended folder structure for a custom Keycloak theme is as follows:
+
+    ├── theme
```
ssilvert, line 13: `theme` can be any string.  It probably wouldn't be called `theme`.  Suggest you change this to something like `my-cool-theme-name`.

```suggestion
    ├── my-cool-theme-name
```

ssilvert (changes requested):

extend-admin-console-quick-start/README.md
agagancarczyk: Hi @ssilvert I have addressed your last comment. Can we merge this now?

PR #413, opened 2023-05-03
Reorganizing directories on a per stack/technology basis

pedroigor:


mposolda (dismissed):

vramik:
Thanks @pedroigor! I like the changes, please note there are some links in the documentation which leads to the quickstarts. I've encountered https://github.com/keycloak/keycloak/blob/main/docs/documentation/server_development/topics/user-storage.adoc?plain=1#L28C28-L32 but there might me more.

pedroigor:
@vramik Thanks. Documentation is already behind and the idea is to review it completely once we are done with the repository. I'll consider those references when updating documentation.

mposolda (approved):

PR #420, opened 2023-05-04
Admin console quickstart update by logo and favicon

agagancarczyk:
Extended quickstart by logo and favicon. Also added pom.xml


ssilvert (changes requested):
You should not need to add your own index.ftl file.

Instead, we need to change the default index.ftl to use `logo`, `logoUrl`, and `favIcon`.

PR #430, opened 2023-05-11
Tmp remove files

pedroigor:


javaee/app-profile-saml-jee-jsp/config/client-import.json
```diff
@@ -1,4 +1,4 @@
-{
+
```
mposolda: This looks like a typo? Parenthesis should not be removed?

mposolda (approved):

javaee/app-profile-saml-jee-jsp/config/client-import.json
pedroigor: Done

PR #431, opened 2023-05-11
Reviewing documentation

pedroigor:
* Review docs for a single quickstart to make sure we are aligned about the format/structure we should be using.

### DIFF
--- a/jakarta/app-authz-jakarta-servlet/README.md
+++ b/jakarta/app-authz-jakarta-servlet/README.md
@@ -5,8 +5,6 @@ Level: Beginner
 Technologies: Jakarta EE  
 Summary: Servlet Application Using Fine-grained Authorization  
 Target Product: <span>Keycloak</span>, <span>WildFly</span>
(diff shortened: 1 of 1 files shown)


jakarta/app-authz-jakarta-servlet/README.md
```diff
@@ -31,26 +29,44 @@ You'll also learn how to use the `AuthorizationContext` object to obtain permiss
...
+* JDK 11
+* Apache Maven 3.8.6
+* Wildfly 28+
+* Keycloak 21+
+* Docker 20+
+
+Starting and Configuring the Keycloak Server
+-------------------
+
+To start a Keycloak Server you can use Docker and just run the following command in the root directory of this quickstart:
```
mposolda, line 43:
```suggestion
To start a Keycloak Server you can use OpenJDK on Bare Metal, Docker, Openshift or any other option described in [Keycloak Getting Started guides]https://www.keycloak.org/guides#getting-started. For example when using Docker just run the following command in the root directory of this quickstart:
```

mposolda (approved):

PR #442, opened 2023-05-19
More improvements

pedroigor:


mposolda (commented):
In the extension/*/README.md file, there are also server startup commands, like for instance:
```
kc.[sh|bat] start-dev --http-port=8180 --http-relative-path="/auth" --spi-events-listener-sysout-exclude-events=CODE_TO_TOKEN,REFRESH_TOKEN
```
Those may be updated as well to remove `--http-relative-path="/auth"` parameter? Or do you prefer to do it in next iteration?

mposolda (approved):

Issue #503, opened 2023-11-03
Automate Keycloak version replacement


bnallapeta:
### Before reporting an issue

- [X] I have read and understood the above terms for submitting issues, and I understand that my issue may be closed without action if I do not follow them.

### Area

docs

### Describe the bug

The current keycloak.yaml file in the Keycloak [getting started](https://www.keycloak.org/getting-started/getting-started-kube) guide for kubernetes contains a placeholder for `$$VERSION$$` that users need to manually replace with the actual version of Keycloak they wish to deploy. This requires manual intervention and could lead to errors or the deployment of outdated versions if not replaced correctly.

### Version

All versions

### Expected behavior

The deployment process should be automated such that the `$$VERSION$$` placeholder in the keycloak.yaml file is dynamically replaced with the latest Keycloak version available from the official releases. This should happen as part of the command used to deploy Keycloak to Kubernetes, without requiring users to manually edit the file.

### Actual behavior

Currently, the user must manually find the latest Keycloak version, edit the keycloak.yaml file, and replace the `$$VERSION$$` placeholder before applying the file with kubectl.

### How to Reproduce?

Check the steps listed in https://www.keycloak.org/getting-started/getting-started-kube

### Anything else?

Proposed https://github.com/keycloak/keycloak/pull/24521 as a potential fix for this issue.

bnallapeta:
@vmuzikar  What should we replace `$$VERSION$$` with? If we simply use the current latest version (21.0.5), we would need to constantly update this file with each new release.

bnallapeta:
I don't see any release specific Github workflow to automate this too. If that was part of the setup, we probably could have added a step to auto-update the version with each release.

vmuzikar:
The `latest` branch is updated automatically (e.g. versions in POMs) with each new release. And looks like we're already doing that, [`set-version.sh`](https://github.com/keycloak/keycloak-quickstarts/blob/bf4b32dd0edd627843c8e48abff067f7ce079bc5/set-version.sh#L7-L8) which is [called](https://github.com/keycloak-rel/keycloak-rel/blob/72a3b5ce0015e0ae5f5d04d0b9e17040a390d733/.github/workflows/x-create-tags.yml#L51) during the release process.

Not sure what's happening there, we need to investigate.

bnallapeta:
@vmuzikar I think its probably because of the lack of escape character? In the PR I had submitted, I have used the escape character for $ and it works. Should I raise a PR with that change?

[...]

PR #504, opened 2023-11-08
Fix sed command to escape $

bnallapeta:
Reason for PR:
- Currently, the [getting started with kubernetes guide](https://www.keycloak.org/getting-started/getting-started-kube) does not have instructions asking the user to replace the placeholder `$$VERSION$$`.
- Following the instructions listed will result in InvalidImageName error on the pod as it will try to pull `quay.io/keycloak/keycloak:$$VERSION$$`
- https://github.com/keycloak/keycloak-quickstarts/issues/503 

PR Description:
- Fixes the sed command in the set-version.sh script to properly escape $ characters.
- Ensures `$$VERSION$$` placeholder in Kubernetes and OpenShift YAML files is correctly replaced with the provided new version.
- Modifies openshift yaml file to use `$$VERSION$$` instead of a hardcoded value.

### DIFF
--- a/openshift/keycloak.yaml
+++ b/openshift/keycloak.yaml
@@ -65,7 +65,7 @@ objects:
                   value: '${KEYCLOAK_ADMIN_PASSWORD}'
                 - name: KC_PROXY
                   value: 'edge'
-              image: quay.io/keycloak/keycloak:21.1.1
+              image: quay.io/keycloak/keycloak:$$VERSION$$
               livenessProbe:
                 failureThreshold: 100
                 httpGet:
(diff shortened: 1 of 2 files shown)


vmuzikar (changes requested):
@bnallapeta Thanks for the PR!

Could you please rebase onto `main` branch? We can't make the changes in the `latest` as it get overwritten and is generated from `main` with each release.

I tried to change the base branch for this PR but it didn't quite work so a rebase is needed.

bnallapeta:
@vmuzikar ah, I didn't notice that it forked the 'latest' branch. Rebased now. 

I also see that a bot has added more changes and set the version in a bunch of other files. Is this expected?

vmuzikar:
Thanks for  the update!

> I also see that a bot has added more changes and set the version in a bunch of other files. Is this expected?

Nope, there should be only one commit – yours.

bnallapeta:
@vmuzikar My bad. Its all good now. Please check.

vmuzikar (approved):
@bnallapeta Works as expected, thank you!

PR #522, opened 2023-12-21
fix typos in readme: links to getting started

nabbisen:
This PR modifies six links to "Keycloak Getting Started guides":

```diff
- [Keycloak Getting Started guides]https://www.keycloak.org/guides#getting-started
+ [Keycloak Getting Started guides](https://www.keycloak.org/guides#getting-started)
```

### DIFF
--- a/extension/action-token-authenticator/pom.xml
+++ b/extension/action-token-authenticator/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>keycloak-quickstart-parent</artifactId>
         <groupId>org.keycloak.quickstarts</groupId>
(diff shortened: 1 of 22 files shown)


mposolda:
@nabbisen There should not be PR sent to the branch `latest` . All PRs to quickstarts should go to branch `main` .

The content of branch `latest` is always overwritten from branch `main` during Keycloak release, so your changes to the branch `latest` would be lost if updates are done only in branch `latest` instead of updating branch `main` . We plan to update quickstarts contributon guide with mentioning this.

nabbisen:
@mposolda 
Thanks so much. Seems to had better to recreate PR.

PR #540, opened 2024-03-13
Move authenticator example to quickstarts

stianst:
PR to remove from keycloak/keycloak: https://github.com/keycloak/keycloak/pull/27850


pedroigor (approved):
Only missing tests, but for me it can be a follow-up.

Issue #551, opened 2024-04-18
ExtendAccountConsoleTest fails on main


hmlnarik:
### Describe the bug

Due to PF5 upgrade, the `ExtendAccountConsoleTest` fails

### Version

main

PR #552, opened 2024-04-18
Fix ExtendAccountConsoleTest

hmlnarik:
Fixes: #551


hmlnarik:
`ExtendAccountConsoleTest` is fixed

PR #569, opened 2024-07-01
Replace KC_URL also in keycloak.json

mhajas:
It seems by default `keycloak.json` response does not have any content-type set and per docs https://github.com/bfncs/string-replace-middleware?tab=readme-ov-file#configuration [string-replace-middleware](https://github.com/bfncs/string-replace-middleware) does not apply string replacement in such case. 

This PR changes the behavior so also `keycloak.json` gets KC_URL replaced and hence can be use the same way as URL in index.html.

### DIFF
--- a/js/spa/app.js
+++ b/js/spa/app.js
@@ -4,9 +4,13 @@ import stringReplace from 'string-replace-middleware';
 const app = express();
 const port = 8080;
 
+const options = {
+  contentTypeFilterRegexp: /.*/,
+}
+
 app.use(stringReplace({
-  KC_URL: process.env.KC_URL || "http://localhost:8180"
-}));
+  "${KC_URL}": process.env.KC_URL || "http://localhost:8180"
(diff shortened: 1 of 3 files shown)


mposolda:
@jonkoops Could you please review? Feel free to close if it is already out-of-context by your latest changes.

jonkoops (changes requested):
As per the [contribution guidelines](https://github.com/keycloak/keycloak-quickstarts/blob/latest/CONTRIBUTING.md) please target the `main` branch and not the `latest` branch.

jonkoops:
@mposolda the changes for the Node.js side might still be relevant, but I believe my recent changes have made the rest no longer relevant.

mhajas:
Thanks for looking into this @jonkoops. I see keycloak.json is no longer used so I am not 100% sure how to fix this now. I think we can close and leave this as is.

PR #578, opened 2024-07-18
Update keycloak-quickstart to work with artifacts from keycloak-clien…

mposolda:
…t repository

closes keycloak/keycloak#31416

This is only draft PR for now as it cannot be merged right now. Few points:
- It requires https://github.com/keycloak/keycloak-client/pull/11 first to be merged in keycloak-client repository
- It uses version `26.0.0-SNAPSHOT` of keycloak-client. This should be either updated to some released version of keycloak-client OR we should make sure that `26.0.0-SNAPSHOT` is available in the snapshot repository, like for instance here https://s01.oss.sonatype.org/content/repositories/snapshots/org/keycloak/keycloak-policy-enforcer/ ,which will probably need setup of nightly builds in keycloak-client repository

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,6 +36,8 @@
 
     <properties>
         <version.keycloak>${project.version}</version.keycloak>
+        <version.keycloak.client>26.0.0-SNAPSHOT</version.keycloak.client>
+
         <servlet.api.30.version>1.0.2.Final</servlet.api.30.version>
         <jboss-jaxrs-api_2.1_spec>1.0.1.Final</jboss-jaxrs-api_2.1_spec>
         <version.jboss.maven.plugin>7.4.Final</version.jboss.maven.plugin>
@@ -77,7 +79,7 @@
             <dependency>
                 <groupId>org.keycloak</groupId>
                 <artifactId>keycloak-policy-enforcer</artifactId>
-                <version>${version.keycloak}</version>
+                <version>${version.keycloak.client}</version>
             </dependency>
             <dependency>
                 <groupId>org.keycloak</groupId>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -36,6 +36,8 @@
 
     <properties>
         <version.keycloak>${project.version}</version.keycloak>
+        <version.keycloak.client>26.0.0-SNAPSHOT</version.keycloak.client>
```
jonkoops, line 39: Why are we using a `26.0.0-SNAPSHOT` version if other repos use the placeholder `999.0.0-SNAPSHOT`? Not saying we should not use 26, but just wondering why we want to use a different convention here.

jonkoops (approved):

pom.xml
mposolda: AFAIK we use `26.0.0-SNAPSHOT` just temporarily to not clash with existing snapshots. 

For example here is already "999" snapshot for policy-enforcer https://s01.oss.sonatype.org/content/repositories/snapshots/org/keycloak/keycloak-policy-enforcer/ , which is available due the snapshot build before this commit in keycloak main: https://github.com/keycloak/keycloak/commit/5526976d1cbf3204f6a2a1ddf189c140d08f47e9 . So now when you setup sonatype snapshot for keycloak-client repository, there will be new "26.0.0" snapshot and it will be clear that you did your sonatype snapshot setup correctly and it is the build from `keycloak-client` repository rather than some older snapshot build from `keycloak` repository :-)

But I believe that we can later update that also to `999.0.0-SNAPSHOT` after some time? @pskopek Do you have any more insight to this?

PR #586, opened 2024-07-29
Fix broken Account Console tests

jonkoops:


extension/extend-admin-console-spi/src/test/java/org/keycloak/quickstart/ExtendAdminConsoleTest.java
```diff
@@ -106,6 +106,8 @@ public void testRealmSettingsAttributes() {
         realmSettingsAttributePage.navigateTo();
         waitForPageToLoad();
 
+        loginPage.login("admin", "admin");
+
```
edewit, line 110:
```suggestion
```
Remove this line this is already done in the previous test method

Issue #593, opened 2024-09-03
Account Console theme no longer customizable starting from version 25.0.0


alaa60:
### Describe the bug

After upgrading from version 24.0.6 to the latest version (25.0.4) of the extend account console quickstart, I noticed that none of the custom styles are being applied. This issue seems to have started from version 25.0.0. Specifically, all styles defined in css/styles.css are not being applied.

### Version

25.0.0

PR #594, opened 2024-09-03
Fix: Update styles.css to comply with PatternFly v5 syntax changes

alaa60:
Fix https://github.com/keycloak/keycloak-quickstarts/issues/593


edewit (changes requested):
You'll have to target `main` instead of `latest`

PR #596, opened 2024-09-10
Move `keycloak-test-helper` module from main repository

### LINKED ISSUE #584: Move keycloak-test-helper to quickstarts repository
### Description

Some context and options are in: https://github.com/keycloak/keycloak/issues/31653 . Agreed with @stianst on the option of moving keycloak-test-helper to quickstarts. This should be fine as keycloak-test-helper is used just in the quickstarts, but it is not used anywhere else.

So this task is about:

- Create dedicated module in the quickstart repository and add https://github.com/keycloak/keycloak/tree/main/misc/keycloak-test-helper there.
- Remove keycloak-test-helper as a module from Keycloak repository. It is used in main repository just in the misc bom, but that BOM will be removed as well (See https://github.com/keycloak/keycloak/issues/31372 ).

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

jonkoops:
Moves the `keycloak-test-helper` module from the main repository to the quickstarts. For the removal of the original module see keycloak/keycloak#32763.

Closes #584


misc/keycloak-quickstarts-test-helper/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+    <packaging>jar</packaging>
+    <dependencies>
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-client-registration-api</artifactId>
+            <version>${version.keycloak}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-admin-client-tests</artifactId>
```
mposolda, line 23: Is it possible to change this to:
```
       <dependency>
            <artifactId>keycloak-admin-client</artifactId>
            <version>${version.keycloak.client}</version>
        </dependency>
```
?

The `keycloak-admin-client-tests` is the admin-client from the `keycloak/keycloak` repository, which is used in the Keycloak testsuite, but we may not want to expose it outside of that. The people should rather use the admin-client from the `keycloak-client` repository in their applications and ideally, that should be also used in quickstarts.

mposolda (commented):
@jonkoops Thanks!

Added only single comment inline.

pom.xml
```diff
@@ -95,17 +100,6 @@
...
             </dependency>
-            <dependency>
-                <groupId>org.keycloak</groupId>
-                <artifactId>keycloak-test-helper</artifactId>
-                <version>${version.keycloak}</version>
-                <exclusions>
-                    <exclusion>
-                        <groupId>org.keycloak</groupId>
-                        <artifactId>keycloak-admin-client-tests</artifactId>
-                    </exclusion>
```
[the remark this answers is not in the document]
mposolda: I've added some comment related to this one here https://github.com/keycloak/keycloak-quickstarts/pull/596#discussion_r1753470326 .

misc/keycloak-quickstarts-test-helper/pom.xml
jonkoops: Makes sense to me, I've incorporated the changes.

PR #598, opened 2024-09-10
Fix various errors in POM files

jonkoops:
Adds some missing information to the POM files that started yielding errors in IntelliJ.


mposolda (approved):
@jonkoops Thanks!

PR #599, opened 2024-09-11
Use Keycloak JS from NPM instead of server

### LINKED ISSUE #600: De-couple quickstarts from statically served Keycloak JS
As part of Keycloak 26 we want to de-couple Keycloak JS from the Keycloak server (see keycloak/keycloak#32820). As part of this effort the quickstarts should no longer use the version of Keycloak JS statically served from the Keycloak server itself, but rather include it through NPM as a regular dependency.

jonkoops:
Replaces the code to load Keycloak JS from the server with the version installed through NPM. This is needed in order to seperate Keycloak JS from the Keycloak server distribution in the future.

Closes #600

### DIFF
--- a/js/spa/app.js
+++ b/js/spa/app.js
@@ -1,14 +1,13 @@
 import express from 'express';
-import stringReplace from 'string-replace-middleware';
 
 const app = express();
 const port = 8080;
 
-app.use(stringReplace({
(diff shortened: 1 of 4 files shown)


mposolda:
@jonkoops @edewit Thanks for the fix and the review!

mposolda (approved):

Issue #622, opened 2024-11-06
 Update OAuth2ResourceServer configuration in Spring Security quickstart


adropofliquid:
### Description

The current Spring Security quickstart uses a deprecated method to configure the OAuth2 resource server. We should update this to use the more modern Lambda DSL approach, which is consistent with other configuration in the quickstart.

### Details

The current configuration uses:

`.oauth2ResourceServer(OAuth2ResourceServerConfigurer::jwt)`

This should be updated to:

```
.oauth2ResourceServer((oauth2) -> oauth2
    .jwt(withDefaults())
)
```

This change aligns with the existing style used for authorizeHttpRequests in the same configuration class.

PR #623, opened 2024-11-06
Issue 622 update springsecurity  oauth config

adropofliquid:
Closes (#622)


mposolda:
@adropofliquid PR looks good, but could you please fix the PR to not contain any version changes etc, but just the change for the spring quickstart, which is mentioned in the GH issue? Ideally there should be only single commit in this PR just with that change.

PR #626, opened 2024-11-26
initial version of the custom admin ui

edewit:
Signed-off-by: Erik Jan de Wit <erikjan.dewit@gmail.com>


stianst (commented):
Is this mechanism of extending the admin console considered a fully supported feature? or rather more preview or even experimental?

If this is considered fully supported, how can we ensure backwards compatibility for custom UIs. From my dummy knowledge of React it seems that it's not likely that we can.

edewit:
As long as we don't change the names of the components that we share, I don't see why we can't

ssilvert (changes requested):
Need to change account references to admin references.

ssilvert (approved):

PR #632, opened 2024-11-29
Create workflow to test quickstarts with nightly keycloak/keycloak-client

rmartinc:
Closes [#35439](https://github.com/keycloak/keycloak/issues/35439)

I have used the sonatype-snapshots repository to do the tests with nightly. The repo is already used in the quickstarts so I think that it's better this way. The scripts use a environment variable `NIGHTLY_TEST` to use nightly versions 999.0.0-SNAPSHOT instead of the current versions defined in the `pom.xml`. The CI file is transformed into a re-usable workflow called `worflow-ci.yml`. This is called by another two workflows: `ci.yml` (which tests current versions defined in `pom.xml`) and `nightly-ci.yml` (it tests the nightly versions). I have setup to run both at pull, similar to what we do in the keycloak-client repo, but we can execute the nightly CI scheduled at night if preferred. I have also changed some versions because they were using fixed versions (now depend on the var, nightly or current). It's a draft for the moment.

### DIFF
--- a/.github/scripts/prepare-local-server.sh
+++ b/.github/scripts/prepare-local-server.sh
@@ -2,13 +2,15 @@
 
 mkdir keycloak-dist
 
-if [[ ( -n "$GITHUB_BASE_REF" &&  "$GITHUB_BASE_REF" == "latest" ) ]] || [[ ( -n "$QUICKSTART_BRANCH" && "$QUICKSTART_BRANCH" != "main" ) ]]; then
-  VERSION=$(grep -oPm1 "(?<=<version>)[^<]+" pom.xml)
-  echo "Using corresponding Keycloak version: $VERSION"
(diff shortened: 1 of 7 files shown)


mposolda (commented):
@rmartinc Thanks! Changes look great. Thanks for fix the references to `server` on all the places, which previously refer to `project.version` .

One point: If I understand correctly, there is also need to update `schedule.yml` and add `nightly-ci.yml` to it? As otherwise the `nightly-ci.yml` won't be executed during nightly build.

rmartinc:
> One point: If I understand correctly, there is also need to update `schedule.yml` and add `nightly-ci.yml` to it? As otherwise the `nightly-ci.yml` won't be executed during quickstarts nightly build.

Yes, you are right. I don't think about the night schedule. I have added a second job to run both workflows in the current schedule. But please try to test this because it doesn't work in my branch because of permissions. So I did it blindly.

mposolda (approved):
@rmartinc

Issue #553, opened 2024-04-19
Where is the old authenticator example?


oculos:
### Describe the bug

According to this, the example should be under extensions:
https://github.com/keycloak/keycloak/commit/1f772d29575e67ca86b30697a2d4088c1cdbd260

I recall seeing it there, but I can't find it anymore. Is it deprecated now? 

### Version

24.0.3

mposolda:
@oculos Update: The PR to quickstarts was here https://github.com/keycloak/keycloak-quickstarts/pull/540 , however it was sent to `keycloak:latest` and hence it was automatically removed during the release of Keycloak 24.0.3 as correct branch for pull requests is `keycloak:main` .

cornelpurcel:
Hi @mposolda , can I try to take care of the PR to introduce the Authenticator example back?

PR #633, opened 2024-12-01
Add Secret Question authenticator back

cornelpurcel:
Closes https://github.com/keycloak/keycloak-quickstarts/issues/553

Attempts to add back the "Authenticator SPI walk through" example that is missing. Right now the reference the docs point to does not exist: https://www.keycloak.org/docs/26.0.6/server_development/#_auth_spi_walkthrough

I just cherrypicked stianst's commit that is missing from the `main` branch and added the implementation of `SecretQuestionRequiredAction.getCredentialType`, as per mposolda's comment: https://github.com/keycloak/keycloak-quickstarts/issues/553#issuecomment-2091269100

Tested it locally with keycloak 26.0.6 and it works fine.
Please let me know if I messed anything up with the squash in terms of history/blame.

### DIFF
--- a/extension/authenticator/README.md
+++ b/extension/authenticator/README.md
@@ -1,0 +1,64 @@
+authenticator: Example custom authenticator
+========================================================
+
+Level: Beginner  
+Summary: Example custom authenticator  
+Target Product: <span>Keycloak</span>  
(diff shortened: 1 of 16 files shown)


mposolda:
@cornelpurcel Thanks

mposolda (approved):

PR #635, opened 2024-12-03
GH workflow for update keycloak-client version

### LINKED ISSUE #631: Create GH workflow in quickstarts for updating client version
This task is about creating GH workflow in keycloak-quickstarts, which will just update the keycloak-client version in the pom.xml. It is for updating this line in the `pom.xml` file: `<version.keycloak.client>26.0.3</version.keycloak.client>`

I suppose that keycloak-client version can be took as parameter of the workflow? The workflow will be manually triggered after each keycloak-client release (we can have follow-up task to trigger it automatically during keycloak-client release).

graziang:
Closes #631

### DIFF
--- a/.github/workflows/update-keycloak-client-version.yml
+++ b/.github/workflows/update-keycloak-client-version.yml
@@ -1,0 +1,55 @@
+name: Update Keycloak Client Version
+
+on:
+  workflow_dispatch:
+    inputs:
+      new_version:
+        description: "New Keycloak client version to update"
+        required: true
+        type: string
+
+jobs:
+  update-keycloak-version:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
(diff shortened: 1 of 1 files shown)


.github/workflows/update-keycloak-client-version.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Commit and push changes
+        if: env.update_needed == 'true'
+        run: |
+          git add pom.xml
+          git commit -m "Update version.keycloak.client to ${{ github.event.inputs.new_version }}"
+          git push origin main
```
mposolda, line 55: Did you try to test if this works? I have some fear if this works as this may require some "Github secret", which does not seem to be set in this repository?

mposolda (approved):
Approving for now. We may fix as a follow-up if push with updated keycloak-client version does not work

PR #637, opened 2024-12-04
Fix workflow for updating client version

### LINKED ISSUE #636: Fix workflow for updating client version
This is follow-up on https://github.com/keycloak/keycloak-quickstarts/issues/631 .

Current client workflow updates the version of keycloak-client and then it tries to push it to keycloak-quickstarts repository. However there is the issue that push fails because keycloak-quickstarts repository does not have the credentials set and hence does not have rights to push.

Example of failed build is here: https://github.com/keycloak/keycloak-quickstarts/actions/runs/12155609969/job/33897517013

graziang:
Closes #636

### DIFF
--- a/.github/workflows/update-keycloak-client-version.yml
+++ b/.github/workflows/update-keycloak-client-version.yml
@@ -53,3 +53,5 @@ jobs:
           git add pom.xml
           git commit -m "Update version.keycloak.client to ${{ github.event.inputs.new_version }}"
           git push origin main
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"


mposolda:
I've tried to push your changes to the branch `test-workflow` and executed it from there, but it failed https://github.com/keycloak/keycloak-quickstarts/actions/runs/12161215480/job/33915287324 . Maybe something different is needed (like using `-f` switch when running `git push`? But not really sure.) 

For the reference: Here similar workflow in keycloak main, which push something on behalf of keycloak-bot https://github.com/keycloak/keycloak/blob/main/.github/workflows/quarkus-next.yml . It's using branch `quarkus-next` instead of `main`, but I guess that is not difference...

mposolda:
PR is no longer needed. It was fine to adjust some branch protection rules.

PR #640, opened 2024-12-04
Updating set-version.sh and adding set-server-version.sh. Updated server version to 26.0.7

### LINKED ISSUE #639: Adjust set-version.sh script in keycloak-quickstarts
This is pre-requisite to https://github.com/keycloak/keycloak/issues/35437 .

The `set-version.sh` script is currently triggered during keycloak-server release (from quickstarts workflow in `keycloak-rel`).

The script currently updates both "project version" (Version of quickstarts themselves) as well as version of the keycloak server. It is needed to update this as during Keycloak server release, we don't want to update version of the project, but just the version of Keycloak server.

mposolda:
closes #639

- Updated `set-version.sh` script to update only project version, but not version of keycloak server

- Introduced `set-server-version.sh` script, which is used to update all the references to keycloak server, but does not update project version. I am planning to run this script from `keycloak-rel` during keycloak server release as during server release, we are not updating anymore the project version, but we need to update "keycloak server version" .

- Small adjustements needed in the `sed` commands as the string `$$VERSION$$` may not be available anymore in the `kubernetes.yml` and `openshift.yml` files. It would be needed to be able to replace also something like `version: 26.0.7` with something like `version: 26.0.8`, which was not supported by previous `sed` scripts, which assume string `$$VERSION$$`

- Updated server version to last released 26.0.7 used separate commit for that to not mix with updated bash scripts. (Hopefully in the future, this will be always done automatically in the quickstarts `main` branch for further keycloak server releases.)

### DIFF
--- a/js/spa/package.json
+++ b/js/spa/package.json
@@ -8,9 +8,9 @@
     "delete-realm": "node scripts/delete-realm.js"
   },
   "dependencies": {
-    "@keycloak/keycloak-admin-client": "https://github.com/keycloak/keycloak/releases/download/nightly/keycloak-admin-client-999.0.0-SNAPSHOT.tgz",
+    "@keycloak/keycloak-admin-client": "26.0.7",
     "express": "^4.18.2",
-    "keycloak-js": "https://github.com/keycloak/keycloak/releases/download/nightly/keycloak-js-999.0.0-SNAPSHOT.tgz"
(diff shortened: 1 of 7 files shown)


graziang (approved):
@mposolda Thanks, LGTM!

mposolda:
@graziang Thanks!

PR #642, opened 2024-12-06
GH workflow for update Node.js adapter version

### LINKED ISSUE #634: Create GH workflow in quickstarts to update node.js version
This is similar task to this one for keycloak-client: https://github.com/keycloak/keycloak-quickstarts/issues/631

We can have workflow in quickstarts, which would be triggered after node.js adapter release. It looks the workflow may need to update this line: https://github.com/keycloak/keycloak-quickstarts/blob/main/nodejs/resource-server/package.json#L13

It may be needed to also make sure that this line is not updated anymore during keycloak-server release. It seems that one is currently triggered from script `set-server-version.sh` , so probably some updates might be needed in that script too.

graziang:
Closes #634

### DIFF
--- a/.github/workflows/update-nodejs-adapter-version.yml
+++ b/.github/workflows/update-nodejs-adapter-version.yml
@@ -1,0 +1,55 @@
+name: Update Node.js Adapter Version
+
+on:
+  workflow_dispatch:
+    inputs:
+      new_version:
+        description: "New Node.js adapter version to update"
+        required: true
+        type: string
+
+jobs:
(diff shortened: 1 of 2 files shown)


mposolda (approved):
@graziang Thanks!

I am merging as it is ok even if https://github.com/keycloak/keycloak-nodejs-connect/issues/520 and https://github.com/keycloak/keycloak/issues/35536 are not yet finished (in case there is Keycloak server release before those are finished, we can just run the Node.JS workflow separately to update the node.js version)

PR #679, opened 2025-04-09
update to latest version

edewit:
fixes: #677
Signed-off-by: Erik Jan de Wit <erikjan.dewit@gmail.com>


Ukalus:
Without this, its basically impossible to create your own console, because the dependency versions are so far back, and i can only guess what breaks after increasing the versions manually,

Issue #684, opened 2025-04-25
Failing CI in javascript quickstart


mposolda:
### Describe the bug

The quickstart nightly builds are failing CI due the javascript quickstart. For example see latest CI build https://github.com/keycloak/keycloak-quickstarts/actions/runs/14656651254 . The error is something like

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@keycloak/keycloak-admin-client' imported from /home/runner/work/keycloak-quickstarts/keycloak-quickstarts/js/spa/scripts/keycloak-admin-client.js
    at new NodeError (node:internal/errors:405:5)
    at packageResolve (node:internal/modules/esm/resolve:916:9)
    at moduleResolve (node:internal/modules/esm/resolve:973:20)
    at defaultResolve (node:internal/modules/esm/resolve:1206:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:404:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:373:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:250:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:39)
    at link (node:internal/modules/esm/module_job:75:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}

Node.js v18.20.8
```

### Version

nightly (from 2025-04-25)

### Expected behavior

Nightly CI working and passing CI for quickstart PRs

### Actual behavior

Nightly CI failing and failing CI for quickstart PRs

PR #689, opened 2025-05-04
Promote best practices for a Kubernetes deployment

### LINKED ISSUE #688: Kubernetes quickstart should use promote practices
### Describe the bug

There are several best practices that are implemented by the Operator that aren't reflected in the quickstart.

### Version

26.2.0

### Expected behavior

The quickstart should promote best practices when deploying on Kubernetes

### Actual behavior

The current quickstart is

* using a Deployment where it should use a StatefulSet
* Using an H2 database while it should show how to use a database
* Doesn't show how to make Kubernetes discovery work
* Doesn't show probes
* Doesn't set resource limits and requests 

### How to Reproduce?

_No response_

### Anything else?

I'll create a PR

ahus1:
Closes #688

(The JavaScript build currently fails due to #684 which is unrelated to this change)

### DIFF
--- a/kubernetes/keycloak.yaml
+++ b/kubernetes/keycloak.yaml
@@ -6,21 +6,41 @@ metadata:
     app: keycloak
 spec:
   ports:
-    - name: http
+    - protocol: TCP
       port: 8080
-      targetPort: 8080
+      targetPort: http
+      name: http
   selector:
     app: keycloak
-  type: LoadBalancer
+  type: ClusterIP
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: keycloak
+  # Used to
+  name: keycloak-discovery
+spec:
+  selector:
+    app: keycloak
+  # Allow not-yet-ready Pods to be visible to ensure the forming of a cluster if Pods come up concurrently
+  publishNotReadyAddresses: true
+  clusterIP: None
+  type: ClusterIP
 ---
 apiVersion: apps/v1
-kind: Deployment
+# Use a stateful setup to ensure that for a rolling update Pods are restarted with a rolling strategy one-by-one.
+# This prevents losing in-memory information stored redundantly in two Pods.
+kind: StatefulSet
 metadata:
   name: keycloak
   labels:
     app: keycloak
 spec:
-  replicas: 1
+  serviceName: keycloak
+  # Run with one replica to save resources, or with two replicas to allow for rolling updates for configuration changes
+  replicas: 2
   selector:
     matchLabels:
       app: keycloak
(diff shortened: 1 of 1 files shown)


kubernetes/keycloak.yaml
```diff
@@ -6,21 +6,41 @@ metadata:
...
+# Use a stateful setup to ensure that for a rolling update Pods are restarted with a rolling strategy one-by-one.
+# This prevents losing in-memory information stored redundantly in two Pods.
+kind: StatefulSet
 metadata:
   name: keycloak
   labels:
     app: keycloak
 spec:
-  replicas: 1
+  serviceName: keycloak
```
shawkins, line 41: We're discussing this in the operator. The docs are not very consistent, but one statement from https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/ is that "StatefulSets currently require a [Headless Service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) to be responsible for the network identity of the Pods. You are responsible for creating this Service."

So this should likely use the discovery service, not the main service.

cc @vmuzikar

kubernetes/keycloak.yaml
```diff
@@ -32,22 +52,127 @@ spec:
       containers:
         - name: keycloak
           image: quay.io/keycloak/keycloak:26.2.2
-          args: ["start-dev"]
+          args: ["start", "-Djgroups.bind.address=$(POD_IP)"]
```
shawkins, line 55: Might be better to put this with JAVA_OPTS_APPEND, since it also relates to jgroups.

shawkins (commented):
Looks good, just a couple of thoughts.

kubernetes/keycloak.yaml
ahus1: I found out the hard way that only the args will do the replacement `$(POD_IP)` - so I don't have a choice here. And this is the way the Operator does it.

shawkins: I was under the impression that you could use these expressions in environment variables also https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config

> And this is the way the Operator does it (probably for the same reason).

I think part of the motivation for that was not wanting to a "standard" environment variable that would cause us to need additional merging logic or release notes if something were already using that in the unsupported pod template.

ahus1: I figured out that re-ordering fixed it: First define `POD_ID`, then define `JAVA_OPTS_APPEND`. 
Applied to next commit.

kubernetes/keycloak.yaml
ahus1: Addressed in next commit.

ahus1:
@shawkins - I applied changes to address your review comments. This should now be ready for a re-review. 
Thanks!

shawkins (approved):
LGTM

PR #704, opened 2025-07-24
use try-with-resources for reading policy-enforcer.json

### LINKED ISSUE #703: use try-with-resources when loading policy-enforcer.json in OAuth2ResourceServerSecurityConfiguration
### Description

The method createPolicyEnforcerFilter() in OAuth2ResourceServerSecurityConfiguration currently reads the policy-enforcer.json file without closing the input stream, which can lead to resource leaks.

### Discussion

_No response_

### Motivation

Using try-with-resources ensures better resource management and avoids potential file handle leaks.

### Details

**Module:** spring/rest-auth-resource-server

**File:** OAuth2ResourceServerSecurityConfiguration

**Problem:** The method uses getResourceAsStream() without a try-with-resources block, so the stream is never explicitly closed.

**Suggested Fix**: Wrap the InputStream in a try-with-resources block to ensure it is properly closed.

tahabozdemir:
wrapped getResourceAsStream() with try-with-resources to prevent resource leak and added null check

Closes #703


mposolda:
@tahabozdemir Could you please rebase on top of latest main?

tahabozdemir:
@mposolda I’ve rebased onto the latest main.

PR #716, opened 2025-10-14
fix(kuberentes): Fix PostgreSQL mount path for PostgreSQL 17

### LINKED ISSUE #715: Incorrect PostgreSQL 17 Mount Point in Quickstart
### Describe the bug

The PostgreSQL [mount point](https://github.com/keycloak/keycloak-quickstarts/blob/b42ed07fca1be50b6fdb4db58ac3b3373e536ece/kubernetes/keycloak.yaml#L160) in the Quickstart manifest kubernetes/keycloak.yaml is incorrect. Simply changing the volume postgres-data to PersistentVolumeClaim will not achieve data persistence.

According to the container image [documentation](https://hub.docker.com/_/postgres), the expected mount point for PostgreSQL 17 is `/var/lib/postgresql/data`.

### Version

26.4.0

### Expected behavior

By simply changing the volume postgres-data from emptyDir to PersistentVolumeClaim, data persistence should be achieved.

### Actual behavior

Even when changing the volume postgres-data from emptyDir to PersistentVolumeClaim, data is lost upon Pod restart.

### How to Reproduce?

Create a PVC and mount it to the PostgreSQL Deployment.

Create a PVC:
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: postgres-pvc
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 5Gi
```

Modify the PostgreSQL section in kubernetes/keycloak.yaml as follows:
```diff
      volumes:
        - name: postgres-data
-         emptyDir: {}
+         persistentVolumeClaim:
+           claimName: postgres-pvc
```

## Proposed Solutions

Either of the following approaches would resolve this issue:

### Option 1: Modify the Mount Point

```diff
            - name: postgres-data
-             mountPath: /var/lib/postgresql
+             mountPath: /var/lib/postgresql/data
```

### Option 2: Explicitly Set PGDATA Environment Variable

```diff
            - name: POSTGRES_LOG_STATEMENT
              value: "all"
+           - name: PGDATA
+             value: /var/lib/postgresql/17/docker
```

makihh:
Close #715

The volume mount path for the PostgreSQL container is `/var/lib/postgresql`, but for PostgreSQL version 17 and earlier container images, the expected mount path is `/var/lib/postgresql/data`. 
https://hub.docker.com/_/postgres

Therefore, I have fixed the mount path to `/var/lib/postgresql/data`.

With this fix, data persistence can be achieved by simply replacing the volume section in the PostgreSQL manifest with a PVC.
```diff
      volumes:
        - name: postgres-data
-         emptyDir: {}
+         persistentVolumeClaim:
+           claimName: postgres-pvc
```

### DIFF
--- a/kubernetes/keycloak.yaml
+++ b/kubernetes/keycloak.yaml
@@ -157,7 +157,7 @@ spec:
           volumeMounts:
             # Using volume mount for PostgreSQL's data folder as it is otherwise not writable
             - name: postgres-data
-              mountPath: /var/lib/postgresql
+              mountPath: /var/lib/postgresql/data
       volumes:
         - name: postgres-data
           emptyDir: {}


tnorimat (approved):
LGTM

mposolda:
@makihh @tnorimat Thanks for the PR and review!

PR #744, opened 2026-04-16
HAProxy with TLS passthrough

### LINKED ISSUE #743: HAProxy with TLS passthrough
### Description

Add a quickstart to show the concept of a reverse proxy with TLS passthrough, using HAProxy

Architecture:
* 2 Keycloak containers + 1 PostgreSQL container in an internal network (no outside access)
* 1 HAProxy in both the internal and external networks. All traffic must come through the proxy.

Configuration/Concepts to show:
* Use the PROXY protocol to forward IP addresses of the clients
* Use the readiness probe in HAProxy
* Use the shutdown delay in Keycloak to align it with the timing of the probe.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

pruivo:
Closes #743

Successful run in my fork: https://github.com/pruivo/keycloak-quickstarts/actions/runs/24525588194/job/71694803268


ahus1 (approved):
Thank you for the PR, works great! I've added the following: 

* Compatibility with selinux (didn't work on my machine)
* Access logs to the Keycloak containers (to see more activity)
* A little playbook on what there is to see when you stop and restart a container
* Moving the containers to the Google mirror, as Docker has a very restrictive license

Happy to hear your thoughts. Not merging yet to wait for other people's feedback.

tkyjovsk:
I tried this on my side as well. In general it works well, and the test script passes without issues, but I hit 2 issues on my machine:

1) With podman rootless networking (I'm using docker commands with podman underneath) KC sees HAProxy's IP instead the client's original IP.
   Trying to debug with wireshark I found that podman's network interfaces are not visible in the system when using rootless mode. It was necessary for me to run the docker commands with `sudo`. Then the client's IP was passed correctly and the NICs were also visible in wireshark.
   Not sure if we want to document this, probably not, but I just wanted to mention it here for reference.

2) In the stop-start scenario the KC container stops and starts again as expected (checking via `docker ps`) but HAproxy still shows it as offline due to L4 connection problem.
   Also, on my side I see the container named `passthrough_keycloak1_1` instead of `passthrough-keycloak1-1`.

pruivo:
The README mentions Docker and Docker Compose. I tried to get an explanation from AI, but it seems as confused as me, suggesting "source IP masquerading - all incoming traffic appears to come from the container gateway" for security reasons or something like "slirp4netns terminates the TCP connection".  

I was never able to reproduce your problem in 2. with Docker :/

ahus1:
> source IP masquerading - all incoming traffic appears to come from the container gateway

I found that the client's IP address is the address of the container gateway (the `.1`), and then HAProxy had `.2`. The IP address `.1` showed up for me in the session overview, and this showed to me that it is working ok.

PR #748, opened 2026-04-22
Updated Traefik passthrough README with health check and configuration details

### LINKED ISSUE #48127: Traefik: Replicate TLS passthrough
### Description

Replicate the functionality for HAProxy functionality with a set of Docker containers, plus add a similar set of docs.

ruchikajha95:
Updated Traefik passthrough README with health check and configuration details

Signed-off-by: Ruchika <ruchika.jha1@ibm.com>

Closes https://github.com/keycloak/keycloak/issues/48127


proxy/traefik/passthrough/haproxy.cfg
```diff
@@ -0,0 +1,32 @@
+global
```
pruivo: can be removed, we don't need haproxy config.

proxy/traefik/passthrough/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+### 4. Check Traefik dashboard
+
+Open http://127.0.0.1:8080/dashboard/ in a browser to verify that Traefik is running and both Keycloak backends are registered as TCP services
+
+### 5. Showcase graceful shutdown
+
+This is a walkthrough through a graceful shutdown of one of the Keycloak instances:
+
+1. Open [http://localhost:8404/stats](http://localhost:8404/stats) in a browser to verify that both Keycloak backends are healthy.
```
pruivo: The link is broken, can you guess why? 😄

pruivo:
We forgot about CI. Please add a job to https://github.com/keycloak/keycloak-quickstarts/blob/1b60f49d7d809599c3cf0d4ca9916dc71c4be715/.github/workflows/workflow-ci.yml

proxy/traefik/passthrough/traefik.yaml
```diff
@@ -0,0 +1,14 @@
...
+api:
+  dashboard: true
+  insecure: true
+
+log:
+  level: INFO
+
+providers:
+  file:
+    directory: "/mnt"
```
ahus1, line 14: It seems to me that this configuration file is missing those settings mentioned in the `README.md`. 

Without those settings, this is not ready-to-start as the HAProxy setup is. 

Please augment the file.

ahus1 (changes requested):
Thank you for the PR. I suppose the traefik file is missing the `services` and `routers` entries. Can you please check?

proxy/traefik/passthrough/traefik.yaml
ruchikajha95: Hello @ahus1  . Thank you so much for the review! The settings you mentioned are intentionally placed in a separate file, keycloak.yaml. Traefik separates its configuration into two distinct files .

To make this clearer , I can add a note to the README explaining the two-file structure and the role each file plays.

ahus1: Thank you, resolved.

ahus1 (changes requested):
I see you previously experimented with a TCP check. What was the result with that? 

BTW, one can set `KC_HTTP_MANAGEMENT_SCHEME: http` so the management endpoint doesn't run HTTPS which might open the door for the TCP based check. Still I wasn't able to get it to work, and traefik didn't provide me with the response it received from Keycloak, so it was hard to debug. 

When in debug mode, I figured out that "\n" was taken as a literal string, and not as a line break. I then use multi-line YAML, which looked better, but then didn't succeed. 

If we go down the route of not having a readiness probe, this should be reflected in the README, possibly with a link to the open issue. 

Without a probe, we would also need to disable `server-async-bootstrap` (this is probably why the tests are failing). 

Without a probe, there is also not much benefit for a graceful shutdown as it is scripted here. We would actually need to have a different playbook for a graceful shutdown where one would first need to update the `keycloak.yaml` to remove one of the servers before shutting it down.

pruivo:
@ahus1 I didn't experiment with HTTP, but the probe in Traefik compares with **all** the server response (looking at the source code).
I'm guessing the expected string would need the HTTP response (headers, status, etc) + Quarkus Health Check JSON response. 

+1 to disable async-bootstrap and +1 to set shutdown-delay to zero.

ahus1:
> the probe in Traefik compares with all the server response (looking at the source code)

Oh, that gives me an idea! See the next push with a suggested fix that works for me locally.

ahus1:
@pruivo / @ruchikajha95 - I've pushed the changes, and the CI is happy :D

Now I wonder if you think this check is good and you want to continue with it and update the docs, or if you think this check is too fragile. 

I'd argue in favor of keeping it, as having a readiness probe gives you lots of benefits like split-brain detection, a node that is too overloaded, can't talk to the database, better shutdown behavior, etc.

pruivo:
@ahus1 It looks fragile (and ugly), mainly because of the charset that may change by the user. But I'm ok to have it in the quickstarts.

ruchikajha95:
> @pruivo / @ruchikajha95 - I've pushed the changes, and the CI is happy :D
> 
> Now I wonder if you think this check is good and you want to continue with it and update the docs, or if you think this check is too fragile.
> 
> I'd argue in favor of keeping it, as having a readiness probe gives you lots of benefits like split-brain detection, a node that is too overloaded, can't talk to the database, better shutdown behavior, etc.

Thank you @ahus1  , @pruivo  .I was experimenting with something similar yesterday but couldn't get it to work on my end !glad it worked now! I'm happy to keep the changes in the quickstart and proceed with the corresponding documentation updates in the Keycloak repo.!

ahus1:
> I'm happy to keep the changes in the quickstart and proceed with the corresponding documentation updates in the Keycloak repo!

Thanks. Yes, please proceed.

proxy/traefik/passthrough/README.md
```diff
@@ -0,0 +1,207 @@
...
+**Traefik health check :**
+
+**Traefik health check:**
+
+Since Traefik operates at the TCP layer for TLS passthrough, it performs TCP-level health checks on each backend server.
+
+```yaml
+        healthCheck:
+          interval: "5s"
+          timeout: "3s"
```
ahus1: After my change to `keycloak.yaml` this is not outdated. Can you please update the README that it matches the configuration? 

It would be good to have a paragraph about the why the health check is chosen, what possible downsides are (fragile, no TLS), and what the alternatives are (simple TCP, different shutdown procedure, no readiness check, no split brain detection).

ahus1 (changes requested):
Thank you for the PR. After my latest push with the health check, the README is unfortunately out of date. Can you please update it? 

Please review if you think any other items would require a longer explanation of "why" and possible alternatives.

pruivo:
@ruchikajha95, the README in `proxy/README.md` needs to be updated too.

proxy/traefik/passthrough/README.md
```diff
@@ -0,0 +1,207 @@
...
+Traefik operates at the TCP layer (Layer 4) and has no visibility into the HTTP content.
+The TLS connection is terminated by Keycloak itself, which means:
+
+- Keycloak holds the TLS certificate and private key, not the proxy.
+- Traefik cannot inspect, modify, or cache HTTP headers or the request body.
+- End-to-end encryption is preserved between the client and Keycloak.
+
+## Architecture
+
+![Architecture diagram](traefikArchitechture.png)
```
pruivo: The image says `TCP passthrough`. Did you mean `TLS passthrough`?

proxy/traefik/passthrough/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+        target: /opt/keycloak/certs
+        type: bind
+        read_only: true
+        bind:
+          selinux: z
+    depends_on:
+      postgres:
+        condition: service_healthy
+    networks:
+      - frontend
```
pruivo: This is wrong; Keycloak must use only the backend network.

proxy/traefik/passthrough/README.md
```diff
@@ -0,0 +1,207 @@
...
+         servers:
+           - address: "keycloak1:8443"
+           - address: "keycloak2:8443"
+```
+- `serversTransport: kc-passthrough` links the load balancer to the transport defined above, enabling PROXY protocol v2 on all backend connections. This requires Keycloak to be configured with KC_PROXY_PROTOCOL_ENABLED=true.
+- Traffic is distributed across both Keycloak instances using round-robin.
+
+**Traefik health check :**
+
+**Traefik health check:**
```
pruivo: duplicated header.

proxy/traefik/passthrough/README.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+Defines the entry point on port 8443 where Traefik accepts incoming TCP/TLS connections.
+
+**Dashboard:**
+```yaml
+api:
+  insecure: true
+```
+- `insecure: true` exposes the Traefik dashboard without any authentication. This is intended for local development and debugging only and must never be used in production.
```
pruivo: It says the same thing twice: 
* `exposes the Traefik dashboard without any authentication` and `Enables the Traefik dashboard on port 8080 without authentication`
* `This is intended for local development and debugging` and `for local testing only`.

proxy/traefik/passthrough/README.md
```diff
@@ -0,0 +1,207 @@
...
+- `HostSNI(*)`matches all TLS connections regardless of the SNI hostname.
+- `priority: 100` ensures this router takes precedence over any other TCP routers that may be defined
+- `tls.passthrough: true` instructs Traefik to forward the raw TLS stream without terminating it. Keycloak handles TLS termination.
+
+**Servers transport with PROXY protocol:**
+```yaml
+       serversTransports:
+       kc-passthrough:
+         proxyProtocol:
+           version: 2
```
pruivo: Indentation is off
```suggestion
  serversTransports:
    kc-passthrough:
      proxyProtocol:
        version: 2
```

proxy/traefik/passthrough/README.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+- Defines a named transport kc-passthrough that enables PROXY protocol v2 on all backend connections. This allows Keycloak to see the original client IP address and requires Keycloak to be configured with KC_PROXY_PROTOCOL_ENABLED=true.
```
pruivo:
```suggestion
- Defines a named transport `kc-passthrough` that enables PROXY protocol v2 on all backend connections. This allows Keycloak to see the original client IP address and requires Keycloak to be configured with KC_PROXY_PROTOCOL_ENABLED=true.
```

proxy/traefik/passthrough/README.md
```diff
@@ -0,0 +1,207 @@
...
+     services:
+     keycloak-service:
+       loadBalancer:
+         serversTransport: kc-passthrough
+         healthCheck:
+           interval: "5s"
+           timeout: "3s"
+         servers:
+           - address: "keycloak1:8443"
+           - address: "keycloak2:8443"
```
pruivo:
```suggestion
  services:
    keycloak-service:
      loadBalancer:
        serversTransport: kc-passthrough
        # Use the JSON syntax to have explicit CR/LF encoding for send/expect.
        healthCheck: {
          interval: "5s",
          timeout: "3s",
          port: 9000,
          send: "HEAD /health/ready HTTP/1.0\r\n\r\n",
          expect: "HTTP/1.0 200 OK\r\ncontent-type: application/json; charset=UTF-8\r\ncache-control: no-store\r\n\r\n"
        }
        servers:
          - address: "keycloak1:8443"
          - address: "keycloak2:8443"
```

proxy/traefik/passthrough/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+Since Traefik operates at the TCP layer for TLS passthrough, it performs TCP-level health checks on each backend server.
+
+```yaml
+        healthCheck:
+          interval: "5s"
+          timeout: "3s"
+```
+
+- Traefik checks each backend every 5s, marking a server as unhealthy if it does not respond within 3s. Unhealthy backends are automatically removed from the load balancer rotation and re-added once they recover
```
pruivo: To avoid duplicating the configuration, I would remove this section and move this line after line 167

Don't forget to mention CLI  option `KC_HTTP_MANAGEMENT_SCHEME`.

proxy/traefik/passthrough/README.md
```diff
@@ -0,0 +1,207 @@
...
+      loadBalancer:
+        serversTransport: kc-passthrough
+        healthCheck:
+          interval: "5s"
+          timeout: "3s"
+        servers:
+          - address: "keycloak1:8443"
+          - address: "keycloak2:8443"
+```
+- Defines the two Keycloak backend servers. Traefik distributes incoming TCP connections across both instances using round-robin load balancing.
```
pruivo: Same here.

proxy/traefik/passthrough/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+- `docker run --rm --network passthrough_backend alpine nc -zv keycloak1 9000`  verify the management port is reachable within the internal Docker network (as seen in the nc test)
```
pruivo: What is this!?

proxy/traefik/passthrough/traefik.yaml
```diff
@@ -0,0 +1,14 @@
+
```
pruivo, line 1: nitpicking, remove empty line

proxy/traefik/passthrough/traefikArchitechture.png
pruivo, line 1: Compress the image or convert to SVG. 5MB is way too big 😃

pruivo (changes requested):

ahus1:
One more thing: Please link to https://github.com/traefik/traefik/pull/12606 to the configuration and the README so people can track this (or upvote it). 

Once HTTP Healthcheck for TCP services becomes available, we would then know to update this.

proxy/traefik/passthrough/README.md
ruchikajha95: This was to verify whether the connection is open or not? If not needed i will remove this

pruivo: not needed

proxy/traefik/passthrough/README.md
ruchikajha95: Modified

proxy/traefik/passthrough/README.md
ruchikajha95: Fixed

ruchikajha95:
@pruivo @ahus1  thanks for the detailed review . I have  inculcated all the comments mentioned above . Please let me know if any further changes are required?

pruivo:
@ruchikajha95 the image is broken.

pruivo (approved):

ahus1 (approved):
Thank you for this PR. I've slightly updated the README to put more emphasis on the situation that we're missing some features from Traefik here. 

The statement "Since Traefik operates at the TCP layer for TLS passthrough, it cannot perform a standard HTTPS health check." is technically correct, but also not something we want to do, as the health probe should always be on a separate port. 

Please review my changes. If you're good with these changes, and @pruivo is also happy, this is then ready to be merged.

pruivo:
@ahus1 I'm good with the changes.

ruchikajha95:
> Thank you for this PR. I've slightly updated the README to put more emphasis on the situation that we're missing some features from Traefik here.
> 
> The statement "Since Traefik operates at the TCP layer for TLS passthrough, it cannot perform a standard HTTPS health check." is technically correct, but also not something we want to do, as the health probe should always be on a separate port.
> 
> Please review my changes. If you're good with these changes, and @pruivo is also happy, this is then ready to be merged.

LGTM as well :) Thanks

PR #763, opened 2026-05-21
Reencrypt shutdown timeout statement fix in Haproxy and Traefik readme 

ruchikajha95:
Reencrypt shutdown timeout statement fix in Haproxy and Traefik readme


ahus1 (approved):
Thank you for the update. I've aligned the compose file with the README.

PR #767, opened 2026-05-26
set-server-version.sh should update server image versions in docker-compose files

### LINKED ISSUE #766: set-server-version.sh should update server image versions in docker-compose files
### Description

The `set-server-version.sh` file is used to update the server image version used throughout the quickstarts at release time. However it does not process the `proxy/**/**/docker-compose.yml` files that were recently added.

We should update the script to ensure that the image is set as expected.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

ryanemerson:
Closes #766

### DIFF
--- a/set-server-version.sh
+++ b/set-server-version.sh
@@ -20,6 +20,8 @@ sed -i 's/ image:\(.*keycloak\):.*/ image:\1:'"$NEW_VERSION"'/g' kubernetes/keyc
 sed -i 's/ version: .*/ version: '"$NEW_VERSION"'/g' openshift/keycloak.yaml
 sed -i 's/ image:\(.*keycloak\):.*/ image:\1:'"$NEW_VERSION"'/g' openshift/keycloak.yaml
 
+find . -name "docker-compose.y*ml" -exec sed -i 's| image:\(.*keycloak/keycloak\):.*| image:\1:'"$NEW_VERSION"'|g' {} +
+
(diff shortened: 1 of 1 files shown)


set-server-version.sh
```diff
@@ -20,6 +20,8 @@ sed -i 's/ image:\(.*keycloak\):.*/ image:\1:'"$NEW_VERSION"'/g' kubernetes/keyc
 sed -i 's/ version: .*/ version: '"$NEW_VERSION"'/g' openshift/keycloak.yaml
 sed -i 's/ image:\(.*keycloak\):.*/ image:\1:'"$NEW_VERSION"'/g' openshift/keycloak.yaml
 
+find . -name "docker-compose.y*ml" -exec sed -i 's| image:\(.*keycloak/keycloak\):.*| image:\1:'"$NEW_VERSION"'|g' {} +
```
pruivo, line 23: Should we skip this step if `NEW_VERSION` is `999.0.0-SNAPSHOT`?

ryanemerson: I don't think it's necessary. We don't check that when updating the k8s/openshift manifests. Looking at the git history and https://github.com/keycloak-rel/keycloak-rel, the automations never call the script with `999.0.0-SNAPSHOT`, it just changes the version to the release in progress.

pruivo (approved):

PR #768, opened 2026-05-26
mTLS to authenticate a client with Traefik reencrypt

### LINKED ISSUE #49210: mTLS to authenticate a client with Traefik reencrypt
### Description

https://www.keycloak.org/server/reverseproxy#_enabling_client_certificate_lookup

ryanemerson:
Closes https://github.com/keycloak/keycloak/issues/49210


proxy/traefik/reencrypt/README.md
```diff
@@ -145,6 +167,12 @@ In Traefik, a `customRequestHeaders` entry with an empty string ("") value remov
           X-Ot-Span-Context: ""
 ```
 
+`X-Forwarded-Tls-Client-Cert` and `X-Forwarded-Tls-Client-Cert-Info` is stripped to prevent an external client from injecting a forged client certificate.
+Traefik is the only component that should set this header, after verifying the certificate presented during the TLS handshake.
+Without this filtering, an attacker could bypass X.509 client certificate authentication by sending a crafted header directly.
```
pruivo: Most likely, Traefik automatically filters those headers, as it does with the X-Forwarded. I'm ok to leave it there for visibility (as we did for the other).

proxy/traefik/reencrypt/README.md
```diff
@@ -159,6 +187,57 @@ Note: If the `KC_PROXY_HEADERS` setting is set to forwarded (see below), Keycloa
...
+    passTLSClientCert:
+      pem: true
+```
+
+Traefik's `passTLSClientCert` middleware with `pem: true` forwards the verified client certificate as a URL-encoded PEM
+string via the `X-Forwarded-Tls-Client-Cert` header. The middleware also removes any existing `X-Forwarded-Tls-Client-Cert`
+header before setting the verified value, providing an additional layer of defense against header spoofing.
+
+The middleware only sets the header when the client presented a valid certificate during the TLS handshake.
+If no certificate was presented, the header is removed entirely.
```
pruivo, line 225: So, did you try this? The filter to remove it may not be required at all.

pruivo (approved):
LGTM.

proxy/traefik/reencrypt/README.md
```diff
@@ -145,6 +167,12 @@ In Traefik, a `customRequestHeaders` entry with an empty string ("") value remov
           X-Ot-Span-Context: ""
 ```
 
+`X-Forwarded-Tls-Client-Cert` and `X-Forwarded-Tls-Client-Cert-Info` is stripped to prevent an external client from injecting a forged client certificate.
+Traefik is the only component that should set this header, after verifying the certificate presented during the TLS handshake.
+Without this filtering, an attacker could bypass X.509 client certificate authentication by sending a crafted header directly.
+
+NOTE: The `X-Forwarded-Tls-Client-Cert` header is hardcoded in the `TraefikProxySslClientCertificateLookup` provider and must match the header set by Traefik's `passTLSClientCert` middleware.
```
ruchikajha95: This statement feels a bit redundant since the same point is made at line 322. I suggest removing this one and keeping the version at line 322, as it is phrased more clearly.

proxy/test-client-cert-lookup.sh
```diff
@@ -0,0 +1,120 @@
...
+    echo ""
+    echo "Arguments:"
+    echo "  hostname     Hostname of the Keycloak cluster"
+    echo "  cacert       Path to the CA certificate (PEM)"
+    echo "  client-key   Path to the client private key (PEM)"
+    echo "  client-cert  Path to the client certificate (PEM)"
+    echo "  subject-dn   Expected subject DN for the X.509 client authenticator"
+    echo ""
+    echo "Example:"
+    echo "  $0 127.0.0.1.nip.io ./certs/haproxy-external/cert.pem ./certs/client/key.pem ./certs/client/cert.pem 'CN=test-client'"
```
ruchikajha95, line 15: Should this be traefik  instead of haproxy?

ruchikajha95 (changes requested):

proxy/traefik/reencrypt/README.md
ryanemerson: The header is removed when no certificate is present and and when an invalid cert is presented.

proxy/traefik/reencrypt/README.md
ryanemerson: Yes it does filter automatically. I'll remove the final sentence.

proxy/test-client-cert-lookup.sh
ryanemerson: This script is used by both the haproxy and Traefik quickstarts and this string is just for example usage. Now that I have rebased on the latest `main`, the file is no longer part of the diff.

ruchikajha95 (approved):

PR #771, opened 2026-06-02
Changes for Traefik reencrypt admin api 

### LINKED ISSUE #49473: Prevent access to the Admin API from external IP addresses for Traefik
### Description

Filter paths based on Keycloak documentation: https://www.keycloak.org/server/reverseproxy#_exposed_path_recommendations

### Value Proposition

.

### Goals

.

### Non-Goals

.

### Discussion

_No response_

### Notes

.

ruchikajha95:
This PR makes the Traefik reencrypt quickstart block external access to Keycloak's Admin API and Admin Console, while keeping the public login paths (/realms/, /resources/ etc)open to everyone. It does this by adding an IP allowlist and splitting traffic into two routers :
-  a public one and an internal one that only allows internal IPs.

**Testing Steps followed**

Start the stack and open the Traefik dashboard at http://127.0.0.1.nip.io:8080/dashboard/ . Under HTTP → Routers you should see both keycloak-public and keycloak-internal, and under Middlewares the ip-allowlist.

Closes keycloak/keycloak#49473
Signed-off-by: Ruchika <ruchika.jha1@ibm.com>

### DIFF
--- a/proxy/traefik/reencrypt/README.md
+++ b/proxy/traefik/reencrypt/README.md
@@ -199,6 +199,7 @@ Because TLS re-encrypt lets Traefik inspect the request path, it can enforce thi
           - "192.168.0.0/16"
           - "10.0.0.0/8"
           - "127.0.0.0/8"
+          - "172.16.0.0/12"
 
   routers:
     # Public router: accessible from any source IP address.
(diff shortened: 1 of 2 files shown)


ahus1:
I've added the range "172.16.0.0/12" as this is also present in HAProxy. And as I needed this to access the admin console once I started it with "localhost"

ahus1 (approved):

ahus1:
@ruchikajha95 - nitpick: When adding a sign-off to the commit message, always add a blank link before it; otherwise it is not correctly picked up in all cases. Thanks!

PR #779, opened 2026-06-15
Fix local CI failure in ArquillianSysoutEventListenerProviderTest

### LINKED ISSUE #778: ArquillianSysoutEventListenerProviderTest.testEventListenerOutput fails locally
### Describe the bug

When running the integration tests locally, the ArquillianSysoutEventListenerProviderTest.testEventListenerOutput test fails. The test asserts that an admin event with operationType=ACTION should be present in the Keycloak server log but it encounters an operationType=UPDATE event instead.

### Version

Nightly

### Expected behavior

The test sequence should align with the operation types generated during the admin console login and navigation. The event listener should capture and verify the expected lifecycle events without throwing a test failure

### Actual behavior

```
[ERROR]   ArquillianSysoutEventListenerProviderTest.testEventListenerOutput:121->assertAdminEvent:210 
Expected: a string containing "ADMIN EVENT: operationType=ACTION" but: was "ADMIN EVENT: operationType=UPDATE
```

### How to Reproduce?

- Open the repository on a machine running macOS
- Run the specific test class using Maven:
```
mvn verify -Pextension  -Dit.test=ArquillianSysoutEventListenerProviderTest
```

### Anything else?

_No response_

uzmamansoor09:
Closes #778.

The ArquillianSysoutEventListenerProviderTest.testEventListenerOutput test was failing locally but passing in GitHub Actions CI due to a race condition with log file reading.       

Added a 2 second delay after registerEventListener() in the @Before setup() method to ensure the UPDATE event is fully written to the log file before clearing it. This prevents the UPDATE event from interfering with subsequent test assertions.

### DIFF
--- a/extension/event-listener-sysout/src/test/java/org/keycloak/quickstart/event/listener/ArquillianSysoutEventListenerProviderTest.java
+++ b/extension/event-listener-sysout/src/test/java/org/keycloak/quickstart/event/listener/ArquillianSysoutEventListenerProviderTest.java
@@ -102,7 +102,6 @@ public class ArquillianSysoutEventListenerProviderTest {
         logReader.start();
 
         registerEventListener();
-        TimeUnit.SECONDS.sleep(2);
     }
 
     @After
--- a/extension/event-listener-sysout/src/test/java/org/keycloak/quickstart/event/listener/ArquillianSysoutEventListenerProviderTest.java
+++ b/extension/event-listener-sysout/src/test/java/org/keycloak/quickstart/event/listener/ArquillianSysoutEventListenerProviderTest.java
@@ -88,7 +88,7 @@ public class ArquillianSysoutEventListenerProviderTest {
     }
 
     @Before
-    public void setup() throws IOException {
+    public void setup() throws IOException, InterruptedException {
         webDriver.manage().timeouts().pageLoadTimeout(30, TimeUnit.SECONDS);
         webDriver.manage().timeouts().implicitlyWait(10, TimeUnit.SECONDS);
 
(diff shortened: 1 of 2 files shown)


rmartinc:
@uzmamansoor09 Most probably this is due to the use of the WatchService on Mac. I remember that the implementation on apple is emulated and very very slow. Can you please check if something simpler like what I did in https://github.com/rmartinc/keycloak-quickstarts/tree/refs/heads/issue-778 (see the last commit) works for you? If it works on Mac, try to replace the watch-service for something similar (if you see I'm just checking the size of the file and waiting the same 2000ms but in 4 steps...).

extension/event-listener-sysout/src/test/java/org/keycloak/quickstart/event/listener/ArquillianSysoutEventListenerProviderTest.java
```diff
@@ -88,7 +88,7 @@ public static void tearDownClass() {
     }
 
     @Before
-    public void setup() throws IOException {
+    public void setup() throws IOException, InterruptedException {
```
rmartinc: I suppose that now this is not needed, right?

rmartinc (changes requested):
Thanks @uzmamansoor09! LGTM! Just a nitpick because I suppose the interruptedException is not needed now. The rest is OK.

extension/event-listener-sysout/src/test/java/org/keycloak/quickstart/event/listener/ArquillianSysoutEventListenerProviderTest.java
uzmamansoor09: Yes, thank you! I've removed it.

rmartinc (approved):
Thanks @uzmamansoor09!

PR #790, opened 2026-09-09
Adding quickstarts for ppolicy for LDAP servers

ahus1:
Related to keycloak/keycloak#52400 and the connected pull request.

### DIFF
--- a/ldap/389ds-ppolicy/Dockerfile
+++ b/ldap/389ds-ppolicy/Dockerfile
@@ -1,0 +1,19 @@
+FROM fedora:42
+
+RUN dnf install -y -q 389-ds-base && dnf clean all
+
+RUN dscreate create-template /tmp/ds.inf \
+    && sed -i \
+        -e 's/;root_password = .*/root_password = admin1234/' \
+        -e 's/;port = .*/port = 3389/' \
+        -e 's/;suffix = .*/suffix = dc=example,dc=org/' \
+        -e 's/;self_sign_cert = .*/self_sign_cert = False/' \
+        -e 's/;create_suffix_entry = .*/create_suffix_entry = True/' \
+        /tmp/ds.inf \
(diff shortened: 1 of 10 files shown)


ahus1:
@ruchikajha95 - I needed to prepare this to test a LDAP issue. As it could be handy in the future and for users, I added it as a quickstart. 

Please review it when you have the time. Happy to join a 1:1 session on this one.

ruchikajha95:
Thank you @ahus1 for taking the time to discuss this PR 1-1. The  context was very helpful, and I agree that having this framework in place will be valuable whenever we need to update policies in the future. I would be requesting  some changes on this PR as discussed :

1. Both setup files (389ds-ppolicy and openldap-ppolicy) try to use the same  port (389).
If a user tries to start both at the same time, the second one will crash because port 389 is already taken.
Worth adding a note to the READMEs (or picking a different host port for one
of the setups, e.g., `3389:3389` for 389ds) so users don't have to debug this.

2. 389ds README: clarify internal vs. external port

3.  set-config-pw.ldif stores config password in cleartext                                                                                                                                                                        
                                                                                                                                                                                                                                   
  The file openldap-ppolicy/set-config-pw.ldif contains olcRootPW: config in plaintext. While acceptable for a test setup (and the README warns about this), CAn we consider using a hashed password ({SSHA}...) for slightly better      
  practice in example code that others may copy.

ahus1:
@ruchikajha95 - I've addressed your comments as follows: 

* The two quickstarts now use distinct ports
* The readmes just name the port, and as they are distinct, there is IMHO no need to talk about that.
* For the quickstart to be more accessible, I'd like to keep it in plain text. Instead, I've updated the file to state that in a production environment the password should be hashed.

ruchikajha95 (approved):

